PR #3, opened 2020-12-06
Update template

eldadfux:
Added explanations as markdown comment, so we can view them during writing and also to make sure they don't appear as part of the spec.

Added a few notes about doc changes and scalability.


christyjacob4 (approved):
LGTM 👍

PR #5, opened 2020-12-26
New RFC: JWT Support for Server API Authentication

eldadfux:


TorstenDittmann:
We probably want to have 2 more endpoints. One to revoke tokens and one for refreshing a token.

For the header, I think we can choose whatever fits best for us 👍

eldadfux:
@TorstenDittmann is refreshing a JWT token not exactly the same as creating a new one? We can allow the creation of a new token using an old (not-yet-expired) token.

Is there even a method to expire a JWT token? according to this the only way is using the TTL:
image

https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6#:~:text=Yes%2C%20the%20tokens%20can%20be,cannot%20do%20it%20on%20demand.&text=Note%3A%20If%20you%20are%20using,expiration%20time%20on%20you%20JWT.

TorstenDittmann:
Must have messed up a bit with OAuth2 tokens. Refresh Endpoint makes no sense at all 👍 

But revoking tokens does have a use case imo, since the tokens are stateless. A major drawback of not maintaining state is that a token is active until it's set expiration date. Therefore even if the user has logged out on the front-end and local storage cleared, anyone one having access to the token can access authenticated routes for that user until the token expires.

We could use Redis to blacklist revoked tokens 🤔 

Some informations: https://yos.io/2016/01/07/stateless-authentication-with-json-web-tokens/#revoking-tokens

eldadfux:
Maybe we can couple the token signature with the user session that created it?

TorstenDittmann:
> Maybe we can couple the token signature with the user session that created it?

That sounds like a good idea, maybe we can offer that as an option when creating a token. That means the expire date and if it should be coupled with the current session.

Edit:

Just realized that would add a ton of logic as well, revoking only the token independent of a session looks like the simpler approach.

christyjacob4:
@TorstenDittmann  @eldadfux  we have some good methods to prevent a jwt from being used by someone else. One of the ways is to validate information about the client at the server end. (Source IP, browser version, OS etc.) 
https://security.stackexchange.com/questions/170599/what-protects-a-jwt-from-being-hijacked-and-used-to-pose-as-the-original-user

Rest looks good @eldadfux

TorstenDittmann:
> @TorstenDittmann @eldadfux we have some good methods to prevent a jwt from being used by someone else. One of the ways is to validate information about the client at the server end. (Source IP, browser version, OS etc.)
> https://security.stackexchange.com/questions/170599/what-protects-a-jwt-from-being-hijacked-and-used-to-pose-as-the-original-user
> 
> Rest looks good @eldadfux

Well, it's kind of the nature of JWT that they can be used anywhere. 

I would just provide the possibility to invalidate those as well.

@eldadfux Beside that, I think it looks good from my side as well 👍

eldadfux:
I'll try to create a POC where the token is only valid as long as the session that created it is still alive and well, if that is not too complex it can be a good security feature. Any ideas for default expiration time? @christyjacob4 @TorstenDittmann

eldadfux:
Added some updates to the RFC from the comments and suggestions you had @TorstenDittmann @christyjacob4

TorstenDittmann:
Now something completely different. Would we get any value introducing scopes to this?

eldadfux:
@TorstenDittmann I think not. The main use-case is to be able to mimic the user behavior on the server-side. Meaning we should use the exact same authorization algorithm we use for the client API. Authentication using JWT will only allow you to view resources the user has read/write access and only call endpoints that are available for an authenticated user.

lohanidamodar:
> I'll try to create a POC where the token is only valid as long as the session that created it is still alive and well, if that is not too complex it can be a good security feature. Any ideas for default expiration time? @christyjacob4 @TorstenDittmann

Up to my knowledge, the JWT is not supposed to be limited to the current session. The nature of JWT is to be valid until it's expiration time or until it's revoked. I might be wrong, but limiting JWT with session doesn't make sense. 

For example, I have a mobile application, I login with user's username and password and save only the JWT for next time.

Then again being a server API auth for server side, might make sense to have session only limitation.

eldadfux:
@lohanidamodar I agree the spec doesn't mention anything about it. This is probably mainly because JWT is designed to be stateless.

Adding the session validation on our side should not have any performance penalty, and because the main use case for it is to extend the Appwrite functionality using a custom server, I think this would not just be a nice to have feature but also an important one.

Torsten shared a resource that discusses the weakness very well. Relying on expiration time feels a bit like a backup plan, and if we can have stronger security, I don't see any reason not to implement it.

Suppose someone needs a long-lived token to manipulate data on his server. In that case, they can use an API key with relevant scopes, which can also be invalidated immediately on the Appwrite console.

idoshamun:
The essence of JWT is to be stateless. It was invented to solve microservices authentication.
By being stateless every service needs only the public key to verify the JWT and read its payload instead of reaching out to a remote server which can be laggy. As you mentioned in the RFC, the best practice is to set the expiration of the JWT to 15 minutes and have a stateful refresh token that is used to refresh the JWT.
If you need to validate the JWT with a database query then you don't need JWT, a random string will do just fine.

PR #4, opened 2020-12-16
New RFC: Response Filters

eldadfux:

PR #6, opened 2020-12-30
New RFC: Appwrite CLI

eldadfux:
Attached RFC #4

### DIFF
--- a/003-jwt-support.md
+++ b/003-jwt-support.md
@@ -1,7 +1,7 @@
 # JWT Support for Server API Authentication <!-- What do you want to call your `awesome_feature`? -->
 
 - Implementation Owner: @eldadfux
-- Start Date: 26-12-2021
+- Start Date: 26-12-2020
 - Target Date: Unknown
 - Appwrite Issue:
   https://github.com/appwrite/appwrite/issues/511
--- a/004-command-line-sdk.md
+++ b/004-command-line-sdk.md
@@ -1,0 +1,160 @@
+# Command Line SDK <!-- What do you want to call your `awesome_feature`? -->
+
+- Implementation Owner: @eldadfux
+- Start Date: 30-12-2020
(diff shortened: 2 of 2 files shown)


TorstenDittmann:
I'm not sure if the PHP approach is the best here, especially due to the dependency on Docker.  Because the CLI is most likely also used in CI/CD, we should keep the runtime as optimal as possible.  For example, Node.js (soon also Deno) has many libraries and resources that allow such a CLI to be generated quite efficiently. It can also be bundled with tools like `pkg` to remove any dependencies at all and supports every major platform.

eldadfux:
The packaging with Docker is designed to comply with our vision of cross-platform / tech agnostic platform. It allows us to have one language-agnostic tool, with minimum requirements to start working with the CLI, and also takes into account that Docker is an already exciting Appwrite dependency. The usage of PHP is obviously leveraging our already in-use tools stack.

Besides the above, Docker also offers our devs easy installation (no pip/npm/brew package manager needed), no language-specific version requirement or incompatibility with older langs, and extra security to the user, as our CLI has no unauthorized access to user operating system.

TorstenDittmann:
Okay, fair and understandable points.  Yesterday I also read an article on the subject which resolved some of my concerns, and I sometimes forget how Docker performs - also in CI / CD.

TorstenDittmann:
For removing the prefix, the only thing that comes to my mind is adding an `Alias` to the shell/term.

christyjacob4:
Looks good @eldadfux  . Shall I go ahead with the implementation? Regarding the Alias, we'll figure out something along the way 👍

christyjacob4 (approved):

eldadfux:
> Looks good @eldadfux . Shall I go ahead with the implementation? Regarding the Alias, we'll figure out something along the way 👍

100%

PR #7, opened 2021-01-11
Feat graphql rfc

eldadfux:


005-graphql-api.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+### Create a New Route
+
+Create a new route called POST /v1/graphql. The new endpoint will initialize the GraphQL parser [webonyx/graphql-php](https://github.com/webonyx/graphql-php). We should use Utopia PHP routes object callbacks and metadata to init GraphQL queries and mutations.
+
+### Get Route by Method Name
```
lohanidamodar, line 58: I don't fully understand what this means. Does it mean that we will be using the same existing routes as GraphQL API actions? I am not sure what it means practically.

Glide7 (approved):

lohanidamodar (commented):
I am still pretty new to GraphQL API,  can we add like a Sample request -> response for Appwrite GraphQL, how it would look?

005-graphql-api.md
eldadfux: It means we'll use the same controllers we use in the REST API as the execution targets for any GraphQL queries we get from end users.

lohanidamodar (approved):

christyjacob4 (approved):
LGTM @eldadfux  . We could add the following links to the references section 
https://tech.okcupid.com/the-shadow-request/

https://tech.okcupid.com/moving-okcupid-from-rest-to-graphql/

TorstenDittmann (approved):
Haven't touched GraphQL yet - based from what I know about it, the RFC looks good 👍🏻

eldadfux:
I'll add some actual examples to give a better idea of the end results.

eldadfux:
I should probably also add a section about the way the Appwrite Database is going to be translated to GraphQL.

PR #8, opened 2021-01-12
Feat Support Heroku

lohanidamodar:


006-adjusting-to-support-heroku.md
```diff
@@ -0,0 +1,126 @@
...
+- There will be an option to switch default storages (uploads, configs, cache and functions) to external services like Amazon S3 or Digitalocean Spaces (may be).
+
+<!-- Please avoid discussing your proposed solution. -->
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+### Introduce new environment variable to disable stats
+Should introduce a new environment variable _APP_STATS, default value is **enabled** and can be set to **disabled** to disable the stats, enabling user to completely drop influxdb/telegraph without any errors
```
eldadfux: Not 100% sure about the env name, we can discuss that in the separate RFC for disabling usage stats.

eldadfux: Don't forget to mention in the docs that using this option, devs should disable both influx, telegraf and the usage worker.

006-adjusting-to-support-heroku.md
```diff
@@ -0,0 +1,126 @@
...
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+<!-- Write your answer below. -->
+
+Not sure what completely disabling the stats mean for a end user, whether or not it's a good idea. Or if there are other ways we can provide such facility.
```
eldadfux: We should also remove the portions of the UI referencing the stats:
- Charts on the project homepage
- Chart on the cloud functions monitoring tab

PR #9, opened 2021-01-12
Ability to Disable usage stats

lohanidamodar:


007-disable-usage-stats.md
```diff
@@ -0,0 +1,118 @@
...
+- Start Date: 12-01-2021
+- Target Date: Unknown
+- Appwrite Issue: https://github.com/appwrite/appwrite/issues/547
+
+## Summary
+
+[summary]: #summary
+
+<!-- Brief explanation of the proposed contribution. Write your answer below. -->
+Due to difficulty in deployment of influxdb and telegraf in services like Heroku, in the Appwrite lite version, usage stats has to be disabled. So, want to introduce a environment variable that will allow disabling usage stats completely there by influxdb and telegraf dependency will no longer be required.
```
eldadfux: and also the Appwrite usage worker that sends the data to telegraf.

007-disable-usage-stats.md
```diff
@@ -0,0 +1,118 @@
...
+
+<!-- Please avoid discussing your proposed solution. -->
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+### Introduce new environment variable
+Introduce a new environment variable to allow disabling usage stats.
+    - _APP_USAGE_STATS = enabled | disabled
```
eldadfux, line 45: The first name might make most sense to me. If we'll have more options in the future we can just use new env var names like: `_APP_USAGE_STATS_OPTIONX` or `_APP_USAGE_STATS_OPTIONY`

eldadfux:
I would also add to the PR the actual steps needed to complete the feature including the references to the places in the controllers and the phtml templates (project home screen and function monitoring tab)

Another thing we should do when usage stats are disabled it to stop sending usage date to the Redis queue in this line: https://github.com/appwrite/appwrite/blob/9c421e2dfc64cc7c929f26e5a9193735e1b1c9c6/app/controllers/general.php#L320

007-disable-usage-stats.md
lohanidamodar: updated.

007-disable-usage-stats.md
lohanidamodar: Updated.

lohanidamodar:
@eldadfux Have made the requested updates.

PR #10, opened 2021-01-13
Feat migration rfc

TorstenDittmann:
Related PR: https://github.com/appwrite/appwrite/pull/821

### DIFF
--- a/008-migration.md
+++ b/008-migration.md
@@ -45,16 +45,25 @@ It will look something like this:
 ```php
 abstract class Migration {
   private function __construct() { }
+
+  public function setProject($project, $projectDB) { /* ... */ }
+  public function forEachDocument($callback) { /* ... */ }
   
   abstract protected function execute();
-  
-  //...
 }
 
 class V07 extends Migration {
-  //...
+    public function execute() { /* ... */ }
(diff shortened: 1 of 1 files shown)


008-migration.md
```diff
@@ -0,0 +1,83 @@
...
+
+class V07 extends Migration {
+    public function execute() { /* ... */ }
+}
+```
+
+Following functions are gonna be added the the main `Appwrite\Migration\Migration` class:
+
+#### `setProject(Document $project, Database $projectDB)`
+Sets the current project for migration. Will be used by the migration script run from the CLI.
```
lohanidamodar, line 61: One thing I don't understand, once a user upgrades to newer version, aren't they suppose to migrate their all the projects, will it work on a project by project basis?

lohanidamodar (commented):
If I understand correctly, this migration is for when users upgrade one version of Appwrite to another right?

008-migration.md
TorstenDittmann: Yes, the `setProject` is just for the class to set namespace & co to know which database. The migration script will loop throug all projects 👍🏻

PR #11, opened 2021-01-22
Feat S3 Support

lohanidamodar:


009-s3-storage-support.md
```diff
@@ -0,0 +1,114 @@
...
+This will add S3 storage adapter as an alternative to Local storage adapter. This will allow users to choose between S3 or Local as the default storage adapter.
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+We need to implement new storage adapters.
+
+### Add S3 adapter
+Add, S3 adapter that implements `Device` and its required functions. Also add more functions that is required for S3. We already have a barebones S3 adapter, where we need to implement the required functions
```
eldadfux: Mention this should be done on utopia-php/storage lib.

009-s3-storage-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+<!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
+
+<!-- Write your answer below. -->
+
+Once we have multiple adapters, we can allow support multiple storage adapters at once, later introduce features like storage bucket. So that Appwrite itself can be used as an interface to multiple storage platforms.
```
eldadfux: S3 support should also allow us to easily extend the new adapter to support DigitalOcean spaces and other S3 compatible services.

009-s3-storage-support.md
```diff
@@ -0,0 +1,207 @@
...
+-->
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+<!-- Write your answer below. -->
+1. Do we support multiple buckets? If yes, how?
```
eldadfux: I would currently support 1 given bucket. Once we add our own 'buckets' feature, each of our bucket could support a different AWS bucket or storage adapter in general (local, DigitalOcean, and others).

009-s3-storage-support.md
```diff
@@ -0,0 +1,207 @@
...
+2. There are options like dual-stack, file versioning, which is yet to be explored, and supporting all those features might be difficult without the SDK, will be much like implementing the whole S3 part of AWS SDK.
```
eldadfux: I would start handling file versioning now, anyway appwrite support only write and delete and no update. But it would be nice to still mention it as a future possibility.

009-s3-storage-support.md
```diff
@@ -0,0 +1,207 @@
...
+3. How do we set Access control ACL
```
eldadfux: Not sure what level of ACL do you refer. Access to the files is already handled on the Appwrite auth layer.

PR #12, opened 2021-02-02
Improving performance for storage adapters

lohanidamodar:


eldadfux:
Sounds good, I would only double check if we have any other stuff to take under consideration with other endpoints or workers. If all is good we can move to implementation. I would avoid merging any of this changes to Appwrite 0.7. We can use an independent branch and release it as soon as we can.

eldadfux:
@lohanidamodar just remembered we should probably also check the createTag endpoint on the cloud functions API

lohanidamodar:
@eldadfux createTag simply uploads, making one external request, there is no call to get mimetype as its fixed. No compression already compressed, no encryption as well so we are safe

lohanidamodar:
@eldadfux to optimize, I can reduce to 1 calls as everything like getting mime-type, file-size,  can be done locally with tmp file before upload even happens. However, the ANTIVIRUS is causing problem here too, 

Is using antivirus to scan tmp file an option, by sharing /tmp directory as volume between appwrite and clamav. This would also resolve issue with external storage as checks is always done locally in /tmp directory. This prevents malicious files from ever getting to storage volume, it reaches only up to tmp directory. 

Not sure if this is a good design choice.

@TorstenDittmann and @christyjacob4 I would appreciate your feedback as regarding this.

TorstenDittmann:
> @eldadfux to optimize, I can reduce to 1 calls as everything like getting mime-type, file-size, can be done locally with tmp file before upload even happens. However, the ANTIVIRUS is causing problem here too,
> 
> Is using antivirus to scan tmp file an option, by sharing /tmp directory as volume between appwrite and clamav. This would also resolve issue with external storage as checks is always done locally in /tmp directory. This prevents malicious files from ever getting to storage volume, it reaches only up to tmp directory.
> 
> Not sure if this is a good design choice.
> 
> @TorstenDittmann and @christyjacob4 I would appreciate your feedback as regarding this.

Any performance concerns with waiting for the upload until AV is finished scanning a big or "complicated" file?

lohanidamodar:
> > @eldadfux to optimize, I can reduce to 1 calls as everything like getting mime-type, file-size, can be done locally with tmp file before upload even happens. However, the ANTIVIRUS is causing problem here too,
> > Is using antivirus to scan tmp file an option, by sharing /tmp directory as volume between appwrite and clamav. This would also resolve issue with external storage as checks is always done locally in /tmp directory. This prevents malicious files from ever getting to storage volume, it reaches only up to tmp directory.
> > Not sure if this is a good design choice.
> > @TorstenDittmann and @christyjacob4 I would appreciate your feedback as regarding this.
> 
> Any performance concerns with waiting for the upload until AV is finished scanning a big or "complicated" file?

@TorstenDittmann the concern waiting for upload for large/complicated file if it exists its already there, because the API only returns after making sure the AV as scanned if I believe correctly @eldadfux ? So making this change should not affect performance.

eldadfux:
This change shouldn't create any problem we don't already have with ClamAV, currently we don't support big files. Most ~100-200mb files should be fine, we'll tackle handling bigger files with separate RFC/PR.

eldadfux:
BTW, We could only scan the files with ClamAV if the /tmp directory would be accessible from both the `appwrite` and `clamav` containers. Today, we only pass ClamAV the file path using TCP communication and ClamAV scans it from the shared uploads volume. Currently we don't have any method to send the file contents over the network instead of the path  + this method will add a big performance penalty on both network + disk read/write and ultimately double or more the time it will take to upload a single file.

lohanidamodar:
@eldadfux Not sure I understand, ClamAV is scanning the file in the Storage follder, we are sharing the same storage folder between them. We are just sending the Path to ClamAV if I understand it correct.

So passing the path to the /tmp directory before upload with it shared to both the containers, shouldn't it result to same outcome?

If I am missing something, and it will not work, for now we can only run the tests on local files for now, and skip if external storage adapters, and let people know in the docs that for external adapters, they can remove the ClamAV container!

TorstenDittmann (approved):

lohanidamodar:
@eldadfux @TorstenDittmann With added support for multi part upload and range read, the extra calls here makes sense and this optimization feels unachievable.
For preview endpoint, as we want to check if the file still exists before serving the cache, the exist call there also makes sense. So I don't see a way of optimizing with the suggestions I posted here with new implementation of storage API with chunked upload support.

PR #14, opened 2021-02-14
RFC AppwriteExceptions on SDKs

lohanidamodar:


eldadfux:
@lohanidamodar please suggest how the swagger format should provide information on the possible errors the API can throw based on the official spec if possible. Some examples would make implementation easier.

Another issue I think we should reference is how to present information about the exceptions in the Appwrite documentation or on the SDKs repo themselves.

TorstenDittmann (approved):

eldadfux (approved):

PR #16, opened 2021-02-15
RFC 010 - Anonymous Login

TorstenDittmann:
First draft of the RFC, will add more technical insight now in terms of constants & co.

Quick link: https://github.com/appwrite/rfc/blob/feat-anon-logion/010-anonymous-login.md

### DIFF
--- a/010-anonymous-login.md
+++ b/010-anonymous-login.md
@@ -11,7 +11,7 @@
 
 This RFC will introduce authentication of anonymous users in Appwrite. These temporary anonymous users can be used to allow users who haven't yet signed up to an app to work with data protected by Appwrite's permission and security rules. If an anonymous user decides to sign up to an app, they can convert their account to an E-Mail or OAuth2 authenticated one, so that they can continue to work with their protected data in future sessions.
 
-## #Problem Statement (Step 1)
+## Problem Statement (Step 1)
 
 [problem-statement]: #problem-statement
 
@@ -45,8 +45,6 @@ The user can convert their account to one, that is authenticated via E-Mail or a
 
 The process for OAuth2 should be very simple, once a user has linked an OAuth2 authentification to an anonymous account - it will be converted.
 
-If there already exists a user with that E-Mail address - the anonymous account will be merged into the already existing one.
-
 For data consistency, the user should be provided the same ID at any point.
 
 ### Endpoints
@@ -63,9 +61,8 @@ This endpoint will be adapted, to behave different on an account that doesn't ha
   - If no account was found with provided E-Mail Adress:
     - Save E-Mail Adress
     - Save Password
-  - If account was found with the provided E-Mail Adress:
-    - Ditch user ID from the anonymous Account and migrate every reference in the Database to the old user ID.
-    - Update current Token to new user ID.
+  - If an account was found:
+    - Throws an error
 
 ### Prior art
 
@@ -78,6 +75,8 @@ Luckily we have a lot of resources and examples around the internet that we can 
 [unresolved-questions]: #unresolved-questions
 
 **What happens to unused anonymous accounts?**
+
+> As commented, this will remain untouched and is in the owners responsibility.
 
 Should we add a TTL for anonymous accounts? Let's say no login to an anonymous user happened in 90 days, we should probably delete that user. This option should be configurable.
 
--- a/010-anonymous-login.md
+++ b/010-anonymous-login.md
@@ -17,7 +17,7 @@ This RFC will introduce authentication of anonymous users in Appwrite. These tem
 
 **What problem are you trying to solve?**
 
-When you develop an application, there might be times when you want to let a user interact with pieces of Appwrite before they're signed in.
+When you develop an application, there might be times when you want to let a user interact with pieces of Appwrite before they're signed in. This also increases the converison rate of applications, since the hurdle of registering is very high.
 
 **What is the context or background in which this problem exists?**
 
--- a/010-anonymous-login.md
+++ b/010-anonymous-login.md
@@ -95,6 +95,10 @@ Write your answer below.
 
 To be fair, I have no idea how to handle this properly. We could convert his data to his old account and also transfer all his stuff to the old user id. But this would break consistency, if there are any references in the database to the anonymous user id.
 
+**What happens to unused anonymous accounts?**
+
+Should we add a TTL for anonymous accounts? Let's say no login to an anonymous user happened in 90 days, we should probably delete that user. This option should be configurable.
+
 **If we add one more way of authentication - it will get messy!**
 
 As mentioned in our meeting before, we might want to decouple the user accounts from email authentication. Meaning, that an account can have multiple ways of authentication. This way we could also disable users from abusing the endpoint of creating an account with an E-Mail address if this is not wanted from the owner of an Appwrite project. 
--- a/010-anonymous-login.md
+++ b/010-anonymous-login.md
@@ -35,12 +35,13 @@ An anonymous user will be treated like a unverified normal user with following l
 
 - No Name
 - No E-Mail address
+- No Password
 
-The user can convert their account to one, that is authenticated via E-Mail or an OAuth2 provider. To convert an account to one that is authenticated with an E-Mail address, we should offer an endpoint similiar to the `/v1/account/verification` one. Meaning, an accouunt can only be converted by clicking a verification link form an E-Mail. This way we ensure an anonymous account doesn't get lost with a typo in the E-Mail address. The flow should look like this:
+The user can convert their account to one, that is authenticated via E-Mail or an OAuth2 provider. To convert an account to one that is authenticated with an E-Mail address, we will use the `PATCH:/v1/account/email` endpoint which requires an email and password for the request. Normally This endpoint is used to update the users E-mail address and requires a password. Since both is not set on an anonymous account, we can use this existing endpoint to save both values to the anonymous account - therefore converting it to a user account. The flow should look like this:
 
-- Anonymous user initiate the E-Mail verification process
-- Verifies his E-Mail via the link provided by the E-mail
-- Will only be converted to a normal user at this point.
+- Anonymous user gets created and uses Appwrite features without providing any personal informations.
+- Passes E-mail adress and password to the `PATCH:/v1/account/email` endpoint.
+- A successful request will convert this account.
 
 The process for OAuth2 should be very simple, once a user has linked an OAuth2 authentification to an anonymous account - it will be converted.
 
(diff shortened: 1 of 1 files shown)


010-anonymous-login.md
```diff
@@ -0,0 +1,114 @@
...
+
+This RFC will introduce authentication of anonymous users in Appwrite. These temporary anonymous users can be used to allow users who haven't yet signed up to an app to work with data protected by Appwrite's permission and security rules. If an anonymous user decides to sign up to an app, they can convert their account to an E-Mail or OAuth2 authenticated one, so that they can continue to work with their protected data in future sessions.
+
+## #Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+When you develop an application, there might be times when you want to let a user interact with pieces of Appwrite before they're signed in.
```
eldadfux: Maybe emphasis that requiring a user to provide an authentication method might also create extra frication and reduce signup conversion rates. This also affect use cases where regulation might require or make you wish to avoid collecting any data about the user.

010-anonymous-login.md
```diff
@@ -0,0 +1,114 @@
...
+- Verifies his E-Mail via the link provided by the E-mail
+- Will only be converted to a normal user at this point.
+
+The process for OAuth2 should be very simple, once a user has linked an OAuth2 authentification to an anonymous account - it will be converted.
+
+For data consistency, the user should be provided the same ID at any point.
+
+### Endpoints
+
+#### `GET: /v1/account/sessions/anonymous`
```
eldadfux: Should be POST

010-anonymous-login.md
```diff
@@ -0,0 +1,114 @@
...
+
+Allow a user to create and login to an anonymous user account.
+
+#### `POST: /v1/account/verification/anonymous`
+
+Allow a user to initiate converting their account to one that is authenticated via E-Mail. This will have the same functionality as the `POST/v1/account/verification` endpoint, except that an E-Mail address needs to be provided in the payload.
```
eldadfux: What happens if in the future will offer account authentication using other methods like SMS? Should we consider allowing anonymous users to use the `PATCH /v1/account/email` using an empty password and then optionally allow app makers to start verification using the existing verification flow?

With this suggestion we can better reuse endpoint and make future verification methods easier to implement

010-anonymous-login.md
```diff
@@ -0,0 +1,114 @@
...
+Write your answer below.
+-->
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+**What happens if the user wants to convert his anonymous account to his already existing user account?**
+
+To be fair, I have no idea how to handle this properly. We could convert his data to his old account and also transfer all his stuff to the old user id. But this would break consistency, if there are any references in the database to the anonymous user id.
```
eldadfux: I would just leave this in the hands of app makers. If they wish, I guess our API flexibly could already allow that kind of "accounts merge" or with minimum intervention from our side.

010-anonymous-login.md
```diff
@@ -0,0 +1,114 @@
...
+
+For example:
+
+I create a new project and in the settings I can choose what authentication methods I want to use. Let's say I want to use E-Mail and Google OAuth2. A user creates an account via the the `POST /v1/account/email` endpoint. This will create an user with that E-Mail and Password as one possible way of logging into this account. Now the user also wants to connect his Google Account via the `GET /v1/account/oauth2/{provider}` endpoint. This endpoint will recognize a already logged in user due to cookies and will add that authentication method to the account. Now all of a sudden the owner needs to add GitHub as an OAuth2 provider to attract more developers. This would also allow multiple E-Mail addresses for a single account.
+
+Also, if we disable E-Mail authentication - this way we can ensure users not creating such accounts over the REST API.
+
```
eldadfux, line 94: Some of this concerns are addressed in my previous comment, at least the ones about multiple auth methods (email + sms).

OAuth login in Appwrite already checks for existing user before creating a new one to avoid multiple users with the same email.

I agree we should allow to disable authentication method from the console. We can adjust the OAuth providers screen to have checkboxes for email/pass and anon login methods at the top. Those settings can be part of the project object.

eldadfux (commented):
Added a few comments and addressed some of the concerns.

010-anonymous-login.md
TorstenDittmann: What do you think about decoupling users from the E-Mail authentication and treat every way fo authentication like OAuth2 providers?

It would end up to something like this:

Bildschirmfoto 2021-02-15 um 11 53 54
Bildschirmfoto 2021-02-15 um 11 50 56

TorstenDittmann: Not saying it needs to be done for this RFC, just something to think about for future implementations.

010-anonymous-login.md
TorstenDittmann: Yes, that should work 👍🏻

010-anonymous-login.md
eldadfux: First screenshot, I think we should just add an attribute to the session and couple between the session and the login type = session type, the is also a plan in the backlog to attach access tokens to the session object.

**Edit**
This might also force us to change the `Devices` tab to `Sessions`.

Second screenshot, let do as part of this feature. I would call the tab `Settings` at least for until we'll have more stuff to split.

Should we turn email/pass and anon login by default for new projects?

TorstenDittmann: Yes, turning them both on by default should be fine 👍🏻 

I'll look into implementing it and adapt the RFC 🙂

010-anonymous-login.md
TorstenDittmann: @eldadfux having some doubts about this now. Anonymous users would need to do 2 API calls (setting email and password) before even being able to initiate this verification flow. We would probably need to have an endpoint to pass the E-Mail and Password to start the verification.

010-anonymous-login.md
TorstenDittmann: Okay 👍🏻

010-anonymous-login.md
TorstenDittmann: @eldadfux some thoughts:

With this we shouldn't use the E-Mail address as an identifier for linking OAuth2 logins to a user account. Especially since this value may not be unique to this user and is not suitable for use as a primary key. For this we should use the `sub` value (or any other unique id to the user) provided by the OAuth2 providers 👍🏻 

Also, maybe the E-Mail address used for the Appwrite user is not the same for the Google Account. This would involve adding a whole new collection, maybe called `authentications` that will be used for authenticating the user and is also refered to in the user's `tokens` for example. This involves heavy refactoring and needs to be discussed and deserves its own RFC.

TorstenDittmann: If we would abstract every authentication method from accounts, we would have a very good starting point for any new authentication method added in the future.

010-anonymous-login.md
eldadfux: Actually this has proved itself very good for us when we separated account creation and login for REST compatibility. As I see it this is a limitation of the REST protocol, but in this case I would favor flexibility over performance, each developer should decide its own app flow, and the 50ms penalty shouldn't be too much fo this kind of process.

Also, creating multiple endpoint just for small performance gains will open the door for us to introduce a lot of added complexity on the Appwrite API.

TorstenDittmann: As discussed, this will be part of the `PATCH: /v1/account/email` endpoint now.

010-anonymous-login.md
```diff
@@ -0,0 +1,102 @@
...
+- Target Date: 16-02-2021
+- Appwrite Issue: https://github.com/appwrite/appwrite/discussions/907
+
+## Summary
+
+[summary]: #summary
+
+This RFC will introduce authentication of anonymous users in Appwrite. These temporary anonymous users can be used to allow users who haven't yet signed up to an app to work with data protected by Appwrite's permission and security rules. If an anonymous user decides to sign up to an app, they can convert their account to an E-Mail or OAuth2 authenticated one, so that they can continue to work with their protected data in future sessions.
+
+## #Problem Statement (Step 1)
```
eldadfux: Extra #

010-anonymous-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+The user can convert their account to one, that is authenticated via E-Mail or an OAuth2 provider. To convert an account to one that is authenticated with an E-Mail address, we will use the `PATCH:/v1/account/email` endpoint which requires an email and password for the request. Normally This endpoint is used to update the users E-mail address and requires a password. Since both is not set on an anonymous account, we can use this existing endpoint to save both values to the anonymous account - therefore converting it to a user account. The flow should look like this:
+
+- Anonymous user gets created and uses Appwrite features without providing any personal informations.
+- Passes E-mail adress and password to the `PATCH:/v1/account/email` endpoint.
+- A successful request will convert this account.
+
+The process for OAuth2 should be very simple, once a user has linked an OAuth2 authentification to an anonymous account - it will be converted.
+
+If there already exists a user with that E-Mail address - the anonymous account will be merged into the already existing one.
```
eldadfux: Should mention this is true only for OAuth2. If another user has this email when using the update email route it will throw an error.

010-anonymous-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### `PATCH: /v1/account/email`
+
+This endpoint will be adapted, to behave different on an account that doesn't have E-Mail and Password set (`null` in this case). If the endpoint turns out to be requested from an anonymous user, it will do the following:
+
+- Checks if the E-Mail address is already registered to an account.
+  - If no account was found with provided E-Mail Adress:
+    - Save E-Mail Adress
+    - Save Password
+  - If account was found with the provided E-Mail Adress:
```
eldadfux: We should throw an error just like it does right now, otherwise any anonymous user can hijack any account he wants because we don't the email address actually belong to him.

010-anonymous-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+Luckily we have a lot of resources and examples around the internet that we can take as an example. I don't think I need to name them here, since a simple Google search will satisfy everyone's curiosity.
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+**What happens to unused anonymous accounts?**
+
+Should we add a TTL for anonymous accounts? Let's say no login to an anonymous user happened in 90 days, we should probably delete that user. This option should be configurable.
```
eldadfux, line 81: I would leave it untouched for now, users can create cloud functions to handle this scenario, this will also raise the question about the user data on deletion.

010-anonymous-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+> This will be covered in a seperate RFC.
+
+As mentioned in our meeting before, we might want to decouple the user accounts from email authentication. Meaning, that an account can have multiple ways of authentication. This way we could also disable users from abusing the endpoint of creating an account with an E-Mail address if this is not wanted from the owner of an Appwrite project. 
+
+For example:
+
+I create a new project and in the settings I can choose what authentication methods I want to use. Let's say I want to use E-Mail and Google OAuth2. A user creates an account via the the `POST /v1/account/email` endpoint. This will create an user with that E-Mail and Password as one possible way of logging into this account. Now the user also wants to connect his Google Account via the `GET /v1/account/oauth2/{provider}` endpoint. This endpoint will recognize a already logged in user due to cookies and will add that authentication method to the account. Now all of a sudden the owner needs to add GitHub as an OAuth2 provider to attract more developers. This would also allow multiple E-Mail addresses for a single account.
+
```
eldadfux, line 92: I'm not sure I fully understand the problem scope, but we can discuss separately.

eldadfux (approved):

010-anonymous-login.md
```diff
@@ -0,0 +1,101 @@
...
+This RFC will introduce authentication of anonymous users in Appwrite. These temporary anonymous users can be used to allow users who haven't yet signed up to an app to work with data protected by Appwrite's permission and security rules. If an anonymous user decides to sign up to an app, they can convert their account to an E-Mail or OAuth2 authenticated one, so that they can continue to work with their protected data in future sessions.
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+When you develop an application, there might be times when you want to let a user interact with pieces of Appwrite before they're signed in. This also increases the converison rate of applications, since the hurdle of registering is very high.
```
christyjacob4, line 20: there might be times when you want to let a user interact with pieces of your application before they're signed in

PR #17, opened 2021-02-23
Realtime RFC

TorstenDittmann:


014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+| collections                | Any update/delete/create events for collections where user has read access |
+| collections.[ID]           | Any update/delete/create events to a given collection where user has read access |
+| collections.[ID].documents | Any update/delete/create events to any document in a given collection where user has read access |
+| documents                  | Any update/delete/create events to documents where user has read access |
+| documents.[ID]             | Any update/delete/create events to a given document where user has read access |
+| files                      | Any update/delete/create events to file where user has read access |
+| files.[ID]                 | Any update/delete/create events to a given file where user has read access |
+| functions                  | Any execution for functions where user has read access       |
+| functions.[ID]             | Any execution event to a given function where user has read access |
+| functions.executions       | Any update to executions where user has read access          |
```
eldadfux: In most cases we have been avoiding a prefix to channel names. `functions.executions` => `executions`

014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+| functions.executions.[ID]  | Any update to a given function execution where user has read access |
+
+### Messages
+
+The message from the server to the clients should be a JSON string and reflect the following object:
+
+| Property  | Description                                                  |
+| --------- | ------------------------------------------------------------ |
+| event     | The name of the event equivalent to the subscribed to channel name. |
```
eldadfux: What do you mean by `equivalent `?

014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+Or as SDK call in JavaScript:
 
+```js
+let sdk = new Appwrite();
+
+sdk
+    .setHostname('appwrite.test')
+    .setProject('5df5acd0d48c2') // Your project ID
+    .setVersion('v1') // Optional and v1 by default
+    .setSecureProtocol(true) // Optional and true by default - will use https/wss or http/ws
```
eldadfux: `setSecureProtocol ` - Not sure about the naming conventions here, don't feel intuitive enough, that said, I can't think of a better name right now. I'll follow back on this if I have an idea.

014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+;
+
+sdk.realtime.subscribe('account', response => {
```
eldadfux: My initial thought was that this can be intuitive on the client scope, like: `sdk.client.subscribe()` or in Flutter: `client.subscribe()`;

014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+});
+
+const documentsUnsubscribe = sdk.realtime.subscribe('documents.XXXXXX', response => {
+  console.log(response); // Callback will be executed on documents.XXXXXX event
+});
+
+documentsUnsubscribe(); // the subscribe() method will return a method to unsubscribe, invalidate the callback and remove the channel
+```
+
+As seen, the initialization of the client SDK changes a lot. Instead of specifying the WebSocket URI, we can work much more flexibly via the hostname and customize the to be used protocols in the future without another breaking change on the client SDKs.
```
eldadfux: This should also be followed up with changed to the Appwrite console, to easily provide values as copy-paste input fields.

I would completely drop the `setVersion` part. We can have the prefix `/v1/` as part of the paths, and reduce a step for better DX. It will also allow us to use both `v1` and `v2` with the same SDK.

014-realtime-api.md
```diff
@@ -90,8 +91,67 @@ Possible protocols to take under future considiration:
...
+
+The client SDK will maintain a single connection and will maintain all the channels, if a channel is added or removed - a new connection will be established with given channels.
```
eldadfux, line 162: Yep, I like this approach. Reduces complexity by a lot, at least for the first phase.

014-realtime-api.md
```diff
@@ -112,24 +172,30 @@ On each new connection, we should validate that the client origin is valid simil
 
 #### Limit payload size
 
+Since we will not support 2-way communication now - we shouldn't allow/handle receiving messages at all for now. Which makes the payload size not a concern for now.
```
eldadfux: Maybe we should response to incoming messages with a "warning" or at least a note?

014-realtime-api.md
```diff
@@ -112,24 +172,30 @@ On each new connection, we should validate that the client origin is valid simil
...
+
+#### Cookie Authentication
+
+On HTTP Handshakes, cookies are usually sent along, which we can use for authentication. If a technology does not provide a native handshake - we can imitate this functionality to handle passing authentication tokens to the endpoint.
+
 #### JWT Authentication (in path / or in message)
 
-Cookies support in WebSocket and other protocols is limited. Using JWT authentication can be easier to implement and pass to the server and more secure as it is valid for 15 minutes.
+Using JWT authentication can be easier to implement and pass to the server and more secure as it is valid for 15 minutes.
```
eldadfux: I would argue JWT is less secure because of that :)

eldadfux (commented):
I would think of adding a note about tests, benchmarks, docs, tutorial, backend integration potential? (Could we make the API key work out of the box like the cookie/jwt auth?)

The changes looks pretty solid to me. Would love to get more feedback for @christyjacob4 @lohanidamodar (especially regarding Flutter specific issues / DX)

lohanidamodar:
@eldadfux @TorstenDittmann The RFC looks good to me, specially the change in SDK to specify hostname, version instead of endpoint seems valid as we plan to support multiple protocols.

014-realtime-api.md
eldadfux: Another though that just came to my mind. A lot of users prefer to skip Traefik, use their own proxy, and switch ports. Theoretically, HTTP and WS can be on different ports :/

014-realtime-api.md
eldadfux: Maybe we can call it `setUnsecure`.

eldadfux: we can also change `setSelfSigned` to `setDevMode`

eldadfux: For ports we can use `setPortForHTTP` and `setPortForWS`

eldadfux: Another options is `setPorts(http, ws)`

014-realtime-api.md
TorstenDittmann: Agree 👍🏻

christyjacob4:
The proposal looks good to me. 

Regarding future support for other protocols like MQTT etc. I believe we must follow an abstract class design like we do for the OAuthAdapters.

eldadfux:
> The proposal looks good to me.
> 
> Regarding future support for other protocols like MQTT etc. I believe we must follow an abstract class design like we do for the OAuthAdapters.

Agree, but I think in this specific case refactoring should be pretty simple, and I wouldn't add it as a requirement before reaching the 2nd protocol.

014-realtime-api.md
TorstenDittmann: I would prefer the very strict way and just disconnect the client. This leaves no room for abuse.

014-realtime-api.md
TorstenDittmann: > For ports we can use `setPortForHTTP` and `setPortForWS`

What about the Port for HTTPS? 🤔 

Maybe we make our life hard without reason by not just passing an endpoint for HTTP and WebSocket.

PR #19, opened 2021-03-01
Response Model RFC

lohanidamodar:


TorstenDittmann (commented):
Looks good to me for now. Can you create a few samples in a language of your choice to demonstrate it technically?

We should also offer an interface for unknown responses like Documents. As you can see here https://appwrite.io/docs/models/documentList, Document is an `any` object right now, if we add an interface to take care of `$id` and all the other stuff we have by default, developers can create their custom responses easily 🙂

TorstenDittmann:
Reading my comment, maybe an abstract class to extend from makes more sense. We will figure something out 👍🏻

014-response-model.md
```diff
@@ -0,0 +1,108 @@
...
+  theoretical background.
+
+This section is intended to encourage you as an author to think about the
+lessons from other software, provide readers of your RFC with a fuller picture.
+If there is no prior art, that is fine - your ideas are interesting to us
+whether they are brand new or if it is an adaptation from other software.
+
+Write your answer below.
+-->
+May popular SDKs for popular softwares, always return proper response objects instead of plain String or JSON. This improves developer experience a lot. A developer can easily understand, what object the API is returning, what methods and properties are available.
```
kodumbeats, line 222: Could we have a reference to a well-implemented SDK with this behavior?

kodumbeats (approved):
🥇

TorstenDittmann:
Can we also cover general Classes in this RFC that we are going to introduce?

For example in TypeScript we would offer something like:
```typescript
// Internal Appwrite Class
class InternalDocument {
    $id!: string;
}

type InternalPermissions = {
    read?: string[],
    write?: string[]
}

// Exported Appwrite Classes
class DatabaseCollection extends InternalDocument {
    $permissions!: InternalPermissions;
}

class DatabaseDocument extends InternalDocument {
    $collection!: string;
    $permissions!: InternalPermissions;
}
```

This way a user can create their models/classes like this:

```typescript
class Movie extends DatabaseDocument {
    title: string;
    year: number;
    published: boolean = false;

    constructor(title: string, year: number) {
        super();
        this.title = title;
        this.year = year;
    }
}
```

lohanidamodar:
## Based On our discussion:
We will be starting with Flutter SDK, implement the Response Model and publish it first. Get it to the MVP and finally move on to other SDKs.

### Blocker:
Separating client and service side response models in Swagger schema

## Things to remember
How ruby and python handles response models if they do?

eldadfux:
Looks good. I would call the base interface just `Document`. I think a `getAttribute`+`setAttribute` methods on the interface/abstract class can be useful for unstructured docs returned from the database (based on custom collections).

TorstenDittmann:
> Looks good. I would call the base interface just `Document`. I think a `getAttribute`+`setAttribute` methods on the interface/abstract class can be useful for unstructured docs returned from the database (based on custom collections).

I think the getters/setters should better be handled by the language itself and the classes should be kept stupid and simple in my opinion. In .NET you wouldn't really need that kind of methods.

Do we wanna introduce a Class or an Interface?
- Interface would force the user to add Appwrite specific properties like `$id`.
- Class would just inherit these properties.

lohanidamodar:
@eldadfux While trying implementing the Response models we got into few questions/issues
1. For base models whether to use Interface or abstract class as pointed above by @TorstenDittmann 
2. Do we create a base Response model that we return from all the services instead of specific object. For example instead of returning `Future<User>` we would return `Future<Response<User>>. This would keep things similar across Services.
3. What do we do with conflicting names as `Locale` service and `Locale` model would conflict. Do we define all models as `AppwriteLocale`, `AppwriteUser` which I think is better approach. In Flutter it was conflicting as all is defined in the same library. We could try to separate the namespace. But would it be possible for every SDK and would it be simple enough?

eldadfux:
> @eldadfux While trying implementing the Response models we got into few questions/issues
> 
> 1. For base models whether to use Interface or abstract class as pointed above by @TorstenDittmann
> 2. Do we create a base Response model that we return from all the services instead of specific object. For example instead of returning `Future<User>` we would return `Future<Response>. This would keep things similar across Services.
> 3. What do we do with conflicting names as `Locale` service and `Locale` model would conflict. Do we define all models as `AppwriteLocale`, `AppwriteUser` which I think is better approach. In Flutter it was conflicting as all is defined in the same library. We could try to separate the namespace. But would it be possible for every SDK and would it be simple enough?

1. I think an abstract class that all models can extend is a must.
2. I think for type safety and IDE autocompletion each method should return the correct model
3. This shouldn't be an issue as long as we properly leverage namespaces, right? We should try and avoid long naming conventions like `AppwriteSomething` they are not intuitive and a bit messy.

PR #21, opened 2021-03-15
016 Database Adapters and Indices

eldadfux:
Parsed version:

https://github.com/appwrite/rfc/blob/910e7e015c7516c643272f3981ee349dc36c5d11/016-database-adapters.md

### DIFF
--- a/016-database-adapters.md
+++ b/016-database-adapters.md
@@ -239,21 +239,21 @@ CREATE TABLE IF NOT EXISTS `documents_[NAME]_authorization` (
 **Collection Attributes** (no update)
 
 * POST /v1/database/collections/:id/attributes (create attribute)
-* GET /v1/database/collections/:id/attributes/:attribute (list attributes)
+* GET /v1/database/collections/:id/attributes (list attributes)
 * GET /v1/database/collections/:id/attributes/:attribute (get attribute)
 * DELETE /v1/database/collections/:id/attributes/:attribute (delete attribute)
 
 **Collection Indexes** (no update)
 
 * POST /v1/database/collections/:id/indexes (create index)
-* GET /v1/database/collections/:id/indexes/:index (list indexes)
+* GET /v1/database/collections/:id/indexes (list indexes)
 * GET /v1/database/collections/:id/indexes/:index (get index)
 * DELETE /v1/database/collections/:id/indexes/:index (delete index)
 
 **Collection Documents**
 
 * POST /v1/database/collections/:id/documents (create document)
-* GET /v1/database/collections/:id/documents/:document (list documents)
+* GET /v1/database/collections/:id/documents (list documents)
 * GET /v1/database/collections/:id/documents/:document (get document)
 * PUT /v1/database/collections/:id/documents/:document (update document)
 * DELETE /v1/database/collections/:id/documents/:document (delete document)
--- a/016-database-adapters.md
+++ b/016-database-adapters.md
@@ -4,7 +4,10 @@
 - Start Date: 07-03-2021
 - Target Date: Unknown
 - Appwrite Issue:
-  [Is this RFC inspired by an issue in appwrite](https://github.com/appwrite/appwrite/issues/)
+  [New database rule types #395](https://github.com/appwrite/appwrite/issues/395)
+  [Optimize default DB indices #506](https://github.com/appwrite/appwrite/issues/506)
+  [Add Fields param in ListDocument and GetDocument APIs #499](https://github.com/appwrite/appwrite/issues/499)
+  [MongoDB in Appwrite #909](https://github.com/appwrite/appwrite/issues/909)
 
 ## Summary
 
--- a/016-database-adapters.md
+++ b/016-database-adapters.md
@@ -141,27 +141,29 @@ Method for quering data:
 * Is Empty
 * Is Not Empty
 
-### Joins / Relationships
+##### Joins / Relationships
 
-## Paging
+#### Paging
 
-Each database adapter should support two methods for paging. The first method is the classic `Limit and Offset`. The second method is `After` paging, which allows better performance at a larger scale.
+Each database adapter should support two methods for paging. The first method is the classic `Limit and Offset`. The second method is `Limit and After` paging, which allows better performance at a larger scale.
 
-## Orders
+#### Orders
 
-Multi-column support, order type, use native types
+Support multi-column orders, multiple order type (ASC, DESC). We should leverage native casting for all types and avoid any manual casting of the data.
 
-## Features
+#### More Features
 
 > Each database collection should hold parallel tables (if needed) with row-level metadata, used to abstract features that are not enabled in all the adapters.
 
-### Row-level Security
+##### Row-level Security
 
-### GEO Queries
+Each database collection will hold metadata information containing all the read and write permissions data for each document in the collection. In SQL based adapters this data can be left joind using a dedicated table postfixed with the name `_permissions`. Row level security could be disabled when accessing the data using admin credentials for improved performance. Security credentials should be indexed by default on every new collection.
 
-### Free Search
+##### GEO Queries
 
-### Filters
+##### Free Search
+
+##### Filters
 
 Allow to apply custom filters on specific pre-chosen fields. Avaliable filters:
 
@@ -169,11 +171,11 @@ Allow to apply custom filters on specific pre-chosen fields. Avaliable filters:
 * JSON (might be redundent with object support)
 * Hashing (md5,bcrypt)
 
-## Caching
+##### Caching
 
 The library should support memory caching using internal or external memory devices for all read operations. Write operations should actively clean or update the cache.
 
-## Encoding
+##### Encoding
 
 All database adapters should support UTF-8 encoding and accept emoji characters.
 
@@ -259,16 +261,22 @@ CREATE TABLE IF NOT EXISTS `documents_[NAME]_authorization` (
 
 ### Documentation
 
+Below is a list of some of the main topics we should cover as part of the database documnetation. This is in addition to the auto-generated API spec.
+
+* Overview (concept, architecture)
 * Data Types
-* Operations
+* Permissions
+* Operations (queries + examples)
+* Attributes (rules) and indexes
 * Security
 * Benchmarks
 * Performance Tips
 * Known Limitations
 
-### Optimization Tools
+### Resources
 
-https://www.eversql.com/
+* SQL Optimization - https://www.eversql.com/
+* Function syntax parsing https://gist.github.com/lamberta/3768814
 
 ### Prior art
 
--- a/016-database-adapters.md
+++ b/016-database-adapters.md
@@ -231,21 +231,29 @@ CREATE TABLE IF NOT EXISTS `documents_[NAME]_authorization` (
 * PUT /v1/database/collections/:id (update collection)
 * DELETE /v1/database/collections/:id (delete collection)
 
-**Collection Attributes** (no delete)
+**Collection Attributes** (no update)
 
 * POST /v1/database/collections/:id/attributes (create attribute)
 * GET /v1/database/collections/:id/attributes/:attribute (list attributes)
 * GET /v1/database/collections/:id/attributes/:attribute (get attribute)
 * DELETE /v1/database/collections/:id/attributes/:attribute (delete attribute)
 
-**Collection Indices/Indexes** (no delete)
+**Collection Indices/Indexes** (no update)
 
 * POST /v1/database/collections/:id/indexes (create index)
 * GET /v1/database/collections/:id/indexes/:index (list indices)
 * GET /v1/database/collections/:id/indexes/:index (get index)
 * DELETE /v1/database/collections/:id/indexes/:index (delete index)
 
-> We'll probably going to use index and indexes for naming conventions, as its is more consistent.
+**Collection Documents**
+
+* POST /v1/database/collections/:id/documents (create document)
+* GET /v1/database/collections/:id/documents/:document (list documents)
+* GET /v1/database/collections/:id/documents/:document (get document)
+* PUT /v1/database/collections/:id/documents/:document (update document)
+* DELETE /v1/database/collections/:id/documents/:document (delete document)
+
+> We'll use index and indexes for naming conventions, as its is more consistent than indexs vs indices.
 
 ### Database Worker
 
--- a/016-database-adapters.md
+++ b/016-database-adapters.md
@@ -12,7 +12,7 @@
 
 <!-- Brief explanation of the proposed contribution. Write your answer below. -->
 
-Appwrite database don't allow a built-in way to customize your DB indices, or under the hood adapters. The idea in this RFC is to make both options available directly from the database dashboard.
+Appwrite's database doesn't have a built-in way to customize DB indexes or under the hood adapters. The idea in this RFC is to make both options available directly from the database dashboard.
 
 ## Problem Statement (Step 1)
 
@@ -22,15 +22,15 @@ Appwrite database don't allow a built-in way to customize your DB indices, or un
 
 <!-- Write your answer below. -->
 
-The current implementation of the Appwrite database is very good and quick to get started easily, but doesn't allow enough flexibilty when your data set grow or when you need to implement more advanced use-cases.
+The current implementation of the Appwrite database is very good and quick to get started easily, but doesn't allow enough flexibility for when your data set grows or when you need to implement more advanced use-cases.
 
 **What is the context or background in which this problem exists?**
 
-Larger data set and more complex projects, requrie more flexibility and transparency regarding the database inner staructure and performance capabilities.
+Larger data set and more complex projects, require more flexibility and transparency regarding the database inner structure and performance capabilities.
 
 **Once the proposal is implemented, how will the system change?**
 
-We will implement simplified data structures that could be manipulated from both the Appwrite dashboard and directly on from the relevant connected DB. We will also add new endpoint to easily add, delete, and view collection rules and indices.
+We will implement simplified data structures that could be manipulated from both the Appwrite dashboard and directly from the relevant connected DB. We will also add new endpoints to easily add, delete, and view collection rules and indexes.
 
 <!-- Write your answer below. -->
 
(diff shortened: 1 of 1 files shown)


kodumbeats:
I'm not fully convinced that a MongoDB adapter makes sense - we'd be abstracting a document store with another document store. Though, I can understand how/why devs end up on the [free-forever Atlas tier](https://developer.mongodb.com/quickstart/free-atlas-cluster/) and why they'd like the compatibility 🤔

lohanidamodar:
Wow in pretty good shape already. MongoDB support with disclosed Limitation should be good. It's been a one of the popular choice as a database for quite some time now.

kodumbeats:
For enterprise support, we should look into supporting ACID transactions:
- [MariaDB](https://mariadb.com/kb/en/transactions/)
- [MySQL](https://dev.mysql.com/doc/refman/8.0/en/commit.html)
- [Postgres](https://www.postgresql.org/docs/current/tutorial-transactions.html)
- [MongoDB](https://docs.mongodb.com/master/core/transactions/)

Another database driver that could be interesting is [sqlite](https://sqlite.org/index.html). Largest advantage is that sqlite databases are single files, which allows for a ton of portability (basically serverless).

016-database-adapters.md
```diff
@@ -0,0 +1,337 @@
...
+* Integer
+* Float
+* Boolean
+* Null
+
+##### Complex Types
+* Array
+* Object
+* Relationships
+  * Reference (collection / document)
```
TorstenDittmann, line 92: This is important and mighty!

Especially since we don't have to nest documents and documents can be referenced in many places.

016-database-adapters.md
```diff
@@ -0,0 +1,337 @@
...
+##### Simple Types
+
+* String
+* Integer
+* Float
+* Boolean
+* Null
+
+##### Complex Types
+* Array
```
TorstenDittmann, line 89: I assume Array can be used with any Type, including array?

016-database-adapters.md
```diff
@@ -0,0 +1,337 @@
...
+
+Our current query syntax is limted and don't allows us to detect given value types, or to use operators like `OR` and `AND`.
+
+**Current**:
+
+`movie.director.name=Michael Bay`
+
+**New Alternative**:
+
+`movie.director.name.equal('Michael Bay')`
```
TorstenDittmann, line 327: Agree, this looks good. We could take a look at Mango Queries for more examples:

**.lt()** Match fields “less than” this one.
**.gt()** Match fields “greater than” this one.
**.lte()** Match fields “less than or equal to” this one.
**.gte()** Match fields “greater than or equal to” this one.
**.eq()** Match fields equal to this one.
**.ne()** Match fields not equal to this one.
**.exists()** True if the field should exist, false otherwise.
**.in()** The document field must be set (optional rules).
**.nin()** The document field must not be set (optional rules).
**.or()** Matches if any of the selectors in the given array match. 
**.nor()** Matches if none of the selectors in the given array match.
**.regex()** A regular expression pattern to match against the document field.

TorstenDittmann (commented):
Already in a good state 👍🏻

016-database-adapters.md
TorstenDittmann: We could also pass them like this over the REST API, which makes it a bit easier to parse:

`?filter[movie.director.name][equal]=Michael+Bay&filter[limit]=3`

christyjacob4:
@eldadfux If we are doing a refactoring, it might be a good idea to include a layer of redis cache as part of Utopia\Database itself. This way, we (and other devs) could use Utopia\Database as a standalone library for their APIs

eldadfux:
> @eldadfux If we are doing a refactoring, it might be a good idea to include a layer of redis cache as part of Utopia\Database itself. This way, we (and other devs) could use Utopia\Database as a standalone library for their APIs

Agree.

christyjacob4:
Things look good @eldadfux . Would like to follow up when a POC is ready which will bring to light more challenges that we haven't thought of yet 😅

christyjacob4 (commented):
Things look good @eldadfux . Would like to follow up when a POC is ready which will bring to light more challenges that we haven't thought of yet 😅

016-database-adapters.md
kodumbeats: For those, like me, who thought Mango was just a misspelled Mongo: 
https://pouchdb.com/guides/mango-queries.html

kodumbeats:
###  Some thoughts about custom indexes:
How will users know which attributes to index? The standard approach is to use the [EXPLAIN statement](https://dev.mysql.com/doc/refman/8.0/en/explain.html) to understand what's happening under the hood, but our abstraction doesn't allow users to write their own queries. Could leverage a product like [EverSQL](https://www.eversql.com/) and design interfaces for selecting proper attributes to index.

### Hosting questions:
#### Where/how do we set storage limits? 
A list of potential limits that other providers use:
- Memory/CPU
- Storage capacity
- Row/document limits
- Frequency/capacity for backups
- Query timeouts
- Region availability / sharding
- Rollbacks

#### References:
- https://www.digitalocean.com/pricing/#managed-databases
- https://elements.heroku.com/addons/heroku-postgresql

TorstenDittmann:
Addendum to the meeting and as a reminder. Topic: Unique Types.

Perhaps these should not necessarily be associated with indexes. Also, should this be done at database or adapter level?

016-database-adapters.md
eldadfux: Not sure, I originally thought to limit it to contain only one primitive type - also to have easier representation in GraphQL.

016-database-adapters.md
eldadfux: > ?filter[movie.director.name][equal]=Michael+Bay&filter[limit]=3

@TorstenDittmann this does look nice, but will be complex to represent in our SDKs.

eldadfux:
> Addendum to the meeting and as a reminder. Topic: Unique Types.
> 
> Perhaps these should not necessarily be associated with indexes. Also, should this be done at database or adapter level?

I think we can leverage the specific adapter implementation for better confidence and performance. Should work fine with all adapters.

016-database-adapters.md
eldadfux: Will it make sense to allow multiple values like this: `movie.director.name.equal('Michael Bay', 'Steven Spielberg')` instead of having an `OR` operator?

kodumbeats:
### Unanswered question
**sqlite3** doesn't use schemas, but instead gets the `schema-name` from the [attached database](https://www.sqlite.org/lang_attach.html) (which is a file, in these cases). What should `Adapter->create()` look like with files instead of schemas?

eldadfux:
> ### Unanswered question
> **sqlite3** doesn't use schemas, but instead gets the `schema-name` from the [attached database](https://www.sqlite.org/lang_attach.html) (which is a file, in these cases). What should `Adapter->create()` look like with files instead of schemas?

A separate file might be a good equivalent

kodumbeats:
Will collections have their own read/write permissions after this refactor?

016-database-adapters.md
```diff
@@ -0,0 +1,340 @@
...
+
+#### Persistency
+
+Each database adapter should support the following actions for fast storing and retrieval of collections of documents.
+
+**Databases** (Schemas for SQL)
+* create
+* delete
+
+**Collections** (Tables for SQL)
```
kodumbeats, line 105: Collections, attributes, and indexes seem to get a $name, and documents have an $id. Will all of these resources get an $id in the background? Or will we just use $name and $id as written?

016-database-adapters.md
```diff
@@ -0,0 +1,340 @@
...
+
+##### Joins / Relationships
+
+#### Paging
+
+Each database adapter should support two methods for paging. The first method is the classic `Limit and Offset`. The second method is `Limit and After` paging, which allows for better performance at a larger scale.
+
+#### Orders
+
+Support multi-column orders, multiple order type (ASC, DESC). We should leverage native casting for all types and avoid any manual casting of the data.
```
lohanidamodar, line 155: Can we support Random order?
This is one of the biggest still existing issue with `Firebase` as well.

016-database-adapters.md
TorstenDittmann: > > ?filter[movie.director.name][equal]=Michael+Bay&filter[limit]=3
> 
> @TorstenDittmann this does look nice, but will be complex to represent in our SDKs.

How so?

For example the Web SDK can expect an object for the filter argument with following structure:

```json
{
   "movie.director.name": {
      "equal": [
         "Michael Bay",
         "Steven Spielberg"
      ]
   },
   "movie.year: {
      "equal": 1989
   }
}
```

This may seem very complex for easy queries, but the filter is in a readable format for computers and the IDE. Multiple values will imply OR condition if possible.

This allows easy manipulation of the filter object and autocomplete in the IDE.

TorstenDittmann: If we allow devs to pass strings like `movie.director.name.equal('Michael Bay')` - there will be no autocomplete since we cannot offer typings for this.

TorstenDittmann: Example can be found [here](https://www.typescriptlang.org/play?#code/KYOwrgtgBA8gDsATgQwC4HtEGcoG8BQURUAogI5jIA2UAvFAOTAXUMA0hxAcuquZTXoMQvfqw7EoAcUTA0SOowDms+YnaciAGWBYsCoVV371HAL758qAJ4JYCFBmwAVW8EUFJAbQAKUAJYg9khomFgAugD8AFxQAEbo6EbIIAA+WKiIgUrpmdle4angEHFIRZCliAX4FlZuUABi-lSoSFiudvSexF5wiOgONrEZWSBK4bHwIU7tbjWWAPQLUFqJANZQABZI7pYAxuggGVAAZs2t2LFNLW0d7l2aUABEEOgAbv7AAHQAJv6yeycXxAyAgwCesW6kiIzAEsS8j2hxCeAFl-HtNshgDQAELIaxPCRI6FPADKrTeoCgpLgnyolSUT0RxHCjzMRKIL3eny+1jkiAheGZRBUcgusQATAAGCUARjZHOerw+31Q-lQRkFUOhSygzkQ1igyDAGAOEDgRlaWx2lkkFjMQA)

eldadfux: It would be very hard to describe this type of un-structured data as part of the REST and GraphQL APIs

TorstenDittmann: This is actually doable and also already done by other tools. For example Directus does a similiar approach:
https://docs.directus.io/reference/api/query/#filter

But I found a blocker for my approach, OpenAPI/Swagger only allows objects in query params without nesting:
https://github.com/OAI/OpenAPI-Specification/issues/1706#issuecomment-458949282

kodumbeats:
**From recent discussion**
We've decided to move forward with a `dot notation + chaining` approach to client filters and discussed URL encoding. We propose the following delimiters:

```
URI
?filter[]=movie.director.name/equal/'Michael Bay'&filter[]=movie.director.gender/equal/'male'

Array
filter = ["movie.director.name/equal/'Michael Bay'", "movie.director.gender/equal/'male'"]
```
**Unresolved questions**
- Should we create a class for `ordering`, so we can pass the object as a param? Something like:
`db.listDocuments($collection, $filter, $order)`

eldadfux:
> **From recent discussion**
> We've decided to move forward with a `dot notation + chaining` approach to client filters and discussed URL encoding. We propose the following delimiters:
> 
> ```
> URI
> ?filter[]=movie.director.name/equal/'Michael Bay'&filter[]=movie.director.gender/equal/'male'
> 
> Array
> filter = ["movie.director.name/equal/'Michael Bay'", "movie.director.gender/equal/'male'"]
> ```
> 
> **Unresolved questions**
> 
> * Should we create a class for `ordering`, so we can pass the object as a param? Something like:
>   `db.listDocuments($collection, $filter, $order)`

Why did we move away from the functions like syntax?

If it is for performance enhancement I think we should take the small penalty for a more intuitive function like syntax. This will also be easier to explain in the docs, and will allow more options in the future like geo location and search "functions".

TorstenDittmann:
> > **From recent discussion**
> > We've decided to move forward with a `dot notation + chaining` approach to client filters and discussed URL encoding. We propose the following delimiters:
> > ```
> > URI
> > ?filter[]=movie.director.name/equal/'Michael Bay'&filter[]=movie.director.gender/equal/'male'
> > 
> > Array
> > filter = ["movie.director.name/equal/'Michael Bay'", "movie.director.gender/equal/'male'"]
> > ```
> > 
> > 
> > **Unresolved questions**
> > 
> > * Should we create a class for `ordering`, so we can pass the object as a param? Something like:
> >   `db.listDocuments($collection, $filter, $order)`
> 
> Why did we move away from the functions like syntax?
> 
> If it is for performance enhancement I think we should take the small penalty for a more intuitive function like syntax. This will also be easier to explain in the docs, and will allow more options in the future like geo location and search "functions".

If you are good with the penalty - either approach is fine with me.

016-database-adapters.md
```diff
@@ -0,0 +1,340 @@
...
+#### Persistency
+
+Each database adapter should support the following actions for fast storing and retrieval of collections of documents.
+
+**Databases** (Schemas for SQL)
+* create
+* delete
+
+**Collections** (Tables for SQL)
+* createCollection($name)
```
kodumbeats, line 106: How should the adapter work with the createCollection API route?
In the PoC, [createCollection](https://github.com/utopia-php/database/blob/v0/src/Database/Database.php#L146-L156) returns a bool, but in the createCollection API route, [createDocument](https://github.com/appwrite/appwrite/blob/master/app/controllers/api/database.php#L58) is used, and a document is returned. I'm not sure how the new createCollection route should look - should the adapter have a `getCollection` method?

kodumbeats:
> > ### Unanswered question
> > **sqlite3** doesn't use schemas, but instead gets the `schema-name` from the [attached database](https://www.sqlite.org/lang_attach.html) (which is a file, in these cases). What should `Adapter->create()` look like with files instead of schemas?
> 
> A separate file might be a good equivalent

I had trouble creating a SQLite adapter due to its differences in PDO. A SQLITE PDO connection is made to a single file database, not just to the RDBMS:

```php
$pdo = new PDO("sqlite:".$filePath);
```

So, after [constructing a SQLite adapter](https://github.com/utopia-php/database/blob/v0/src/Database/Adapter/MariaDB.php#L30), the `create()` abstract method's behavior can't be replicated.

Do you have any thoughts/suggestions?

016-database-adapters.md
eldadfux: I think we have 2 options, either have a get collection, or we can store the collection metadata as a document. Although the names are the same, this are 2 separate layers that don't have to act exactly the same.

eldadfux:
> > > ### Unanswered question
> > > **sqlite3** doesn't use schemas, but instead gets the `schema-name` from the [attached database](https://www.sqlite.org/lang_attach.html) (which is a file, in these cases). What should `Adapter->create()` look like with files instead of schemas?
> > 
> > 
> > A separate file might be a good equivalent
> 
> I had trouble creating a SQLite adapter due to its differences in PDO. A SQLITE PDO connection is made to a single file database, not just to the RDBMS:
> 
> ```
> $pdo = new PDO("sqlite:".$filePath);
> ```
> 
> So, after [constructing a SQLite adapter](https://github.com/utopia-php/database/blob/v0/src/Database/Adapter/MariaDB.php#L30), the `create()` abstract method's behavior can't be replicated.
> 
> Do you have any thoughts/suggestions?

We should strive to reach a situation where all the differences between different adapters are handled in the constructor method. You can pass the path, store it in the instance and use it later.

lohanidamodar:
@eldadfux @kodumbeats @TorstenDittmann Anyone thinking about Database Transactions during refactoring? Can we support those?

lohanidamodar:
@eldadfux @kodumbeats Support for server side timestamps. Like document creation, update timestamps. may be option for auto filled server side timestamps. 🤔

lohanidamodar:
@eldadfux @kodumbeats Pointer 
Can we add support for updating Array elements:
https://firebase.google.com/docs/firestore/manage-data/add-data#update_elements_in_an_array

Can we add support for updating nested fields (if we support JSON type) https://firebase.google.com/docs/firestore/manage-data/add-data#update_fields_in_nested_objects

JaviBonilla:
> @eldadfux @kodumbeats @TorstenDittmann Anyone thinking about Database Transactions during refactoring? Can we support those?

Transactions (set of read / write atomic operations) would be definetively great to prevent inconsistency issues in the database

016-database-adapters.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Utopia Database
+
+We'll implement an extra database abstraction layer as part of the Utopia project. This layer will include support for multiple database, and will expose a consistent API which will act as the lowest common denominator for all the supported adapters.
+
+The list of supported adapters will include:
+* Postgres
+* MySQL
+* MariaDB
+* MongoDB¹
```
alexweininger, line 74: When it comes time to write documentation for these changes, I think it would be really nice to have a section outlining the differences between the supported databases. 

Although I know by design they will all function the same, there might be subtle differences that developers might want to know.

Developers using Appwrite will most likely care about:
- Disk size of database
- Portability, (ease of backup/restore)
- Speed
- Resource cost (CPU, memory, etc)

JaviBonilla: Just a comment for your consideration. Are you considering aggregation functions? Such as count, sum, max, min, etc. of a particular rule in the collection, this could be quite useful.

https://en.m.wikipedia.org/wiki/Aggregate_function

eldadfux: Probably not in step one, but I can definitely see the value here as we build the future roadmap of this API

fightforlife:
Will the relationships part of the query operations be filled with things like AND and OR?
https://github.com/appwrite/rfc/blob/910e7e015c7516c643272f3981ee349dc36c5d11/016-database-adapters.md#joins--relationships
Relevant Issue:
https://github.com/appwrite/appwrite/issues/1174

kodumbeats:
Hi @fightforlife 👋🏻 

Our idea for `relationships` is to have a way to associate one collection with another, and we're still working on the design/technical implementation of this feature. 

Our current filters work as a list of AND statements, and our refactor adds support for OR statements. This is what the syntax will look like:
https://gist.github.com/kodumbeats/83df7d970096db9e7ba998669e554dee

alexweininger:
Can a collection have more than one index?

eldadfux:
> Can a collection have more than one index?

Yes. We are currently following the under the hood db limits for max number of indexes. 

For MariaDB: ~64
For MySQL: ~64
For MongoDB: ~64

PR #25, opened 2021-05-26
017 - Storage Buckets

TorstenDittmann:
https://github.com/appwrite/rfc/blob/feat-storage-buckets/017-storage-buckets.md

### DIFF
--- a/017-storage-buckets.md
+++ b/017-storage-buckets.md
@@ -193,6 +193,9 @@ whether they are brand new or if it is an adaptation from other software.
 
 Write your answer below.
 -->
+- [About Storage Buckets](https://docs.uipath.com/orchestrator/docs/about-storage-buckets)
+- [Google Cloud Storage](https://cloud.google.com/storage/docs/json_api/v1/buckets)
+- [Google Cloud Storage - Key Terms](https://cloud.google.com/storage/docs/key-terms)
 
 ### Unresolved questions
 
@@ -222,4 +225,6 @@ Possible solutions can be the following:
 
 <!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
 
-<!-- Write your answer below. -->
+- The ability to switch to a different Adapter with existing files - meaning all files will be transfered.
+- The ability to move files between Buckets
+- The ability to change the encryption setting with existing files - meaning all files can be converted.
--- a/017-storage-buckets.md
+++ b/017-storage-buckets.md
@@ -1,0 +1,96 @@
+# Title <!-- What do you want to call your `awesome_feature`? -->
+
+- Implementation Owner: (your github @username)
+- Start Date: (today's date, dd-mm-yyyy)
+- Target Date: (expected date of completion, dd-mm-yyyy)
+- Appwrite Issue:
+  [Is this RFC inspired by an issue in appwrite](https://github.com/appwrite/appwrite/issues/)
+
+## Summary
+
+[summary]: #summary
+
+<!-- Brief explanation of the proposed contribution. Write your answer below. -->
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+<!-- Write your answer below. -->
+
+**What is the context or background in which this problem exists?**
+
+<!-- Write your answer below. -->
+
+**Once the proposal is implemented, how will the system change?**
+
+<!-- Write your answer below. -->
+
+<!-- Please avoid discussing your proposed solution. -->
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+<!--
+This is the technical portion of the RFC. Explain the design in sufficient detail keeping in mind the following:
+
+- Its interaction with other parts of the system is clear
+- It is reasonably clear how the contribution would be implemented
+- Dependencies on libraries, tools, projects or work that isn't yet complete
+- New API routes that need to be created or modifications to the existing routes (if needed)
+- Any breaking changes and ways in which we can ensure backward compatibility.
+- Use Cases
+- Goals
+- Deliverables
+- Changes to documentation
+- Ways to scale the solution
+
+Ensure that you include examples, code-snippets etc. to allow the community to understand the proposed solution. **It would be best if the examples use naming conventions that you intend to use during the actual implementation so that changes can be suggested early on during the development.**
+
+Write your answer below.
+
+-->
+
+### Prior art
+
+[prior-art]: #prior-art
+
+<!--
+
+Discuss prior art, both the good and the bad, in relation to this proposal. A
+few examples of what this can include are:
+
+- Does this functionality exist in other software and what experience has their
+  community had?
+- For other teams: What lessons can we learn from what other communities have
+  done here?
+- Papers: Are there any published papers or great posts that discuss this? If
+  you have some relevant papers to refer to, this can serve as a more detailed
+  theoretical background.
+
+This section is intended to encourage you as an author to think about the
+lessons from other software, provide readers of your RFC with a fuller picture.
+If there is no prior art, that is fine - your ideas are interesting to us
+whether they are brand new or if it is an adaptation from other software.
+
+Write your answer below.
+-->
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+<!-- Write your answer below. -->
+
+### Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+<!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
+
+<!-- Write your answer below. -->
--- a/017-storage-buckets.md
+++ b/017-storage-buckets.md
@@ -77,7 +77,7 @@ A Bucket will have following Model:
   $write: string[];
   name: string;
   enabled: boolean;
-  adapter: 'local'|'s3'
+  adapter: 'local'|'s3'|'etc'
   encrypted: boolean;
   antivirus: boolean;
   adapterConfig?: object;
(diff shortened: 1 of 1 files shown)


017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+- **Maximum File Size Limit**
+  - This setting decides the maximum single file size that can be uploaded.
+- **Maximum Number of Files**
+  - This setting decides how many files a Bucket can contain.
+- **Accepted File Extensions**
+  - This setting decides what file extensions can be uploaded.
+- **Encryption**
+  - This setting decides if files are encrypted. This will come into play, when encrypting large files causes too much load. This is not allowed to be changed after a file has been added to the bucket.
+- **Virus Scan**
+  - This setting decides if files are scanned by ClamAV after upload. Same reasoning ass **Encryption**.
```
lohanidamodar:
```suggestion
  - This setting decides if files are scanned by ClamAV after upload. Same reasoning as **Encryption**.
```

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+#### Prevent Breaking Changes
+
+To prevent a huge breaking change in the Storage Services API's - we are allowing users to use the current Storage implementation as well. Now there is the problem of offering a consistent API pattern across the services.
+
+Let's take a look at the `GET: /v1/storage/files/{fileId}` endpoint. This will result into retrieving a file from the current Storage service. Now if a file is inside a Bucket - the endpoint would have to look like `GET: /v1/storage/files/{bucketId}/{fileId}`. This becomes problematic, since right now it is not possible to allow optional Parameters in the Path with the same Controller/SDK Method and would require refactoring in all SDKs, Utopia and Appwrite.
```
lohanidamodar, line 225: 1. I like the endpoint `/v1/storage/files/{bucketId}/{fileId}
2. If no bucket is created may be appwrite can handle the bucket as `default` internally
3. Internally any path passed only as `/v1/storage/files/{fileId}` appwrite internally converts to `/v1/storage/files/default/{fileId}`

lohanidamodar:
May be not real time but I'would also love to have and be able to get the usage via API
1. Be able to get `total space used`
2. Be able to get `total bandwidth transferred (upload and download)

This is useful if,
for example I create separate bucket for each of my users and limit their max bandwidth and max storage.

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+  - This setting decides the maximum single file size that can be uploaded.
+- **Maximum Number of Files**
+  - This setting decides how many files a Bucket can contain.
+- **Accepted File Extensions**
+  - This setting decides what file extensions can be uploaded.
+- **Encryption**
+  - This setting decides if files are encrypted. This will come into play, when encrypting large files causes too much load. This is not allowed to be changed after a file has been added to the bucket.
+- **Virus Scan**
+  - This setting decides if files are scanned by ClamAV after upload. Same reasoning ass **Encryption**.
+- **TTL**
```
eldadfux, line 70: We should make sure we calculate storage usage properly when TTL expires

eldadfux: Should we also allow TTL per file?

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+  maximumFileSize?: number;
+  allowedFileExtensions?: string;
+}
+```
+
+### New Endpoints
+
+In this section I will explain all necessary endpoints for integratting Storge Buckets.
+
+#### `GET: /v1/storage/bucket/`
```
eldadfux: In order to be persistent with our current conventions this should be `/v1/storage/buckets/`

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+
+This endpoint returns a list of all Files in a specific Bucket. *Server & Client*
+
+Payload:
+- **filters = []** Array of filter strings.
+- **limit = 25** Maximum number of Buckets to return.
+- **offset = 0** Offset value.
+- **search = ''** Search query.
+
+#### `POST: /v1/storage/bucket`
```
eldadfux: To simplify the API, and avoid passing complex object I think we should seriously consider have an individual POST endpoint for each adapter. Each endpoint will be able to receive primitive params that the adapter needs. Internally we can store the data as an object. A unique index for a unique adapter will make the documentation and API easier to understand and will be more consistent with current behavior that favorite the usage of plain primitive params.

For example:
``POST: /v1/storage/bucket/s3`
basic params for bucket creation + 
string - accessKey
string- secretKey
string - bucketName
string - region
string - aclFlag

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+
+Payload:
+- **name = null** Bucket name.
+- **write = null** An array of strings with write permissions.
+- **encrypted = null** Enable Encryption.
+- **antivirus = null** Enable Anti Virus.
+- **adapter = null** Adapter to use.
+- **adapterConfig = null** Adapter Config.
+- **ttl = null** TTL for files.
+- **maximumFiles = null** Maximum amount of files.
```
eldadfux, line 157: What happens when we are over capacity after update?

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+- **maximumFileSize = null** Maximum file size.
+- **allowedFileExtensions = null** Allowed File extensions.
+
+#### `DELETE: /v1/storage/bucket/{bucketId}`
+
+This endpoint deletes a Bucket by its unique ID. *Server*
```
eldadfux, line 163: This endpoint should trigger a background delete of all files in the bucket from both db and storage. This action should probably be delayed by x time to avoid abuse or actions by unauthorized users. The delay will give a time window to revert the action on the server admin level.

eldadfux: We might want to do something similar on a collection delete

017-storage-buckets.md
```diff
@@ -0,0 +1,230 @@
...
+
+Possible solutions can be the following:
+
+- Add the targeted Bucket as a query param to `GET: /v1/storage/files/default/{fileId}?bucket={bucketId}`.
+  - This breaks our API Standards - but could be a good tradeoff in terms of the migration for develoeprs, since there will be no breaking change.
+- Adding a keyword passing to the Bucket parameter that falls back to the default Storage `GET: /v1/storage/files/default/{fileId}`.
+  - This can be done without introducing any breaking changes to the SDK usage
+- Allowing to do calls to `GET: /v1/storage/files/YYYY/XXXX` and `GET: /v1/storage/files//XXXX`
+  - Not a 100% sure this is compliant with any REST Standards
+- Having 2 separate endpoints for each, the default Storage and Storage Buckets, with its own SDK methods
```
eldadfux: `Having 2 separate endpoints for each, the default Storage and Storage Buckets, with its own SDK methods`

I think this is the best and simplest option for us. We can change the names of old endpoints in the new SDK to have a prefix we decide.

017-storage-buckets.md
TorstenDittmann: Same with the user limits - we won't allow any more uploads.

017-storage-buckets.md
TorstenDittmann: Added 👍🏻

017-storage-buckets.md
TorstenDittmann: Added.

017-storage-buckets.md
TorstenDittmann: Added.

eldadfux: We should also add and emphasis the **delayed** part, if a malicious user is triggering a full bucket delete we want to have a timeframe where we can still stop the operation.

017-storage-buckets.md
eldadfux: We will do something like that **temporarily** for backward compatibility, but this will not be consisted with the way we handle all other endpoint paths across Appwrite ([service]/[entity]/[id]).

017-storage-buckets.md
```diff
@@ -0,0 +1,240 @@
...
+
+Additional Payload for **Local** (`/v1/storage/buckets/local`):
+- *none*
+
+Additional Payload for **S3** (`/v1/storage/buckets/s3`):
+- **adapterAccessKeyId** S3 Access Key ID.
+- **adapterSecretKey** S3 Secret Key.
+- **adapterBucket** S3 Bucket Name.
+- **adapterRegion** S3 Region.
+- **adapterAcl** S3 Access Control List.
```
eldadfux:
```suggestion
- **accessKeyId** S3 Access Key ID.
- **secretKey** S3 Secret Key.
- **bucket** S3 Bucket Name.
- **region** S3 Region.
- **acl** S3 Access Control List.
```

The adapter context is already clear from the path/title

017-storage-buckets.md
```diff
@@ -0,0 +1,240 @@
...
+
+Payload:
+- **name** Bucket name.
+- **write** An array of strings with write permissions.
+- **encrypted = true** Enable Encryption.
+- **antivirus = true** Enable Anti Virus.
+- **ttl = false** Enables TTL for files.
+- **maximumFiles = null** Maximum amount of files.
+- **maximumFileSize = null** Maximum file size.
+- **allowedFileExtensions = null** Allowed File extensions.
```
eldadfux:
```suggestion
- **name** Bucket name.
- **write** An array of strings with write permissions.
- **encrypted = true** Enable Encryption.
- **antivirus = true** Enable Anti Virus.
- **ttl = false** Enables TTL for files.
- **maxFiles = null** Maximum amount of files.
- **maxFileSize = null** Maximum file size.
- **fileExtensions = null** Allowed File extensions. Leave empty to allow any extension.
```

Added a few suggestions to param names - still not 100% sure about them.

I think we should consider to implement bot `maxFiles` and `ttl` options in a later stage, not sure how well the DB will scale with them.

eldadfux (changes requested):

eldadfux (approved):

PR #26, opened 2021-06-25
Custom IDs RFC

lohanidamodar:


018-custom-id-support.md
```diff
@@ -0,0 +1,128 @@
...
+- Collections
+- Documents
+- Buckets
+- Files
+- Users
+- Function tags
+- Tasks
+- Webhooks
+
+The create resource endpoints for above resources will have an optional **id** parameter where users can send their custom id following the guidelines for validation. In every endpont, if the **id** parameter is empty, Appwrite will still generate the ids as it is already doing.
```
TorstenDittmann: I think the custom id will be a required parameter.

018-custom-id-support.md
```diff
@@ -0,0 +1,128 @@
...
+- Can be at most 32 characters
+- Can be Alphanumeric
+- Can contain `_` but not at the begining of the id
+- Cannot contain any other special characters
+
+### Initial implementation
+
+The first implementation will not contain following features:
+- TTL
+- Maximum Number of files
```
TorstenDittmann: Is this a leftover from the Storage Bucket RFC? 😄

018-custom-id-support.md
```diff
@@ -0,0 +1,128 @@
...
+- Tasks
+- Webhooks
+
+The create resource endpoints for above resources will have a first required **id** parameter where users can send their custom id following the guidelines for validation. In every endpont, if the **id** parameter is received empty (required for backward compatibility with existing SDKs) or pre defined string, Appwrite will still generate the ids as it is already doing.
+
+We can use the existing key validator to validate the custom id as well.
+
+### Some guidelines for custom ids
+
+- Can be at most 32 characters
```
TorstenDittmann: We maybe should allow 36 characters to support GUID's 🤔

018-custom-id-support.md
lohanidamodar: Ah thanks. removed

018-custom-id-support.md
eldadfux: @kodumbeats your feedback please 🙏🏻

018-custom-id-support.md
```diff
@@ -0,0 +1,122 @@
...
+
+Write your answer below.
+-->
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+Projects and Users being sensitive may be we can exclude them with custom ids.
```
eldadfux: Not sure if we care about users ids being possible, the only problematic thing with a project is and ID called `console`.

018-custom-id-support.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### SDKs
+How to make SDKs look pretty when people don’t want custom **id** and want Appwrite to generate 
+- We could pass predefined String "uid()" -> `database.createCollection("uid()", "Avatars")`
```
eldadfux: Or `unique()` to be more verbose, and keep our style of trying to avoid abbreviation where possible.

018-custom-id-support.md
kodumbeats: According to my research, we should be safe so long as we stay under 64 characters, pretty standard across SQL backends:
- https://mariadb.com/kb/en/identifier-names/#maximum-length
- https://dev.mysql.com/doc/mysql-reslimits-excerpt/5.6/en/identifier-length.html
- https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
- https://docs.mongodb.com/manual/reference/limits/#naming-restrictions

lohanidamodar: @TorstenDittmann @kodumbeats updated to make it 36 chars

018-custom-id-support.md
lohanidamodar: done!

Meldiron:
- This list should also include functions, right? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R54
- Is customID case-sensitive?
- Is there any way to achieve auto-completition on front-end since we have this? For example if I create collection "unfinishedOrders", can my IDE auto-complete me so I dont end-up typing "unfiinshedOrders" by mistake?
- 36 is long enough, but since clients can create it, maaabye some day someone will use dynamic data in document/collection ID, such as "chatBetweenUser8491568495AndUser879489465". Could this limit be editable in ENV? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R71
- What happens if I try to create function that already exists, or maybe even tag? It will throw error or update it? Update seems to be more friendly solution but may break a lot of current setups.

lohanidamodar:
> * This list should also include functions, right? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R54
> * Is customID case-sensitive?
> * Is there any way to achieve auto-completition on front-end since we have this? For example if I create collection "unfinishedOrders", can my IDE auto-complete me so I dont end-up typing "unfiinshedOrders" by mistake?
> * 36 is long enough, but since clients can create it, maaabye some day someone will use dynamic data in document/collection ID, such as "chatBetweenUser8491568495AndUser879489465". Could this limit be editable in ENV? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R71
> * What happens if I try to create function that already exists, or maybe even tag? It will throw error or update it? Update seems to be more friendly solution but may break a lot of current setups.

@Meldiron nice insight.

- I should include Functions
- I believe IDs will be case sensitive
- I don't think auto completion will be possible at the moment, can't figure out in my head how it would work
- I think we want to limit the id so that we can validate, keep it secure
- If id already exists it should throw and error, don't like the idea of updating it, might be friendly but error prone

eldadfux:
> > * This list should also include functions, right? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R54
> > * Is customID case-sensitive?
> > * Is there any way to achieve auto-completition on front-end since we have this? For example if I create collection "unfinishedOrders", can my IDE auto-complete me so I dont end-up typing "unfiinshedOrders" by mistake?
> > * 36 is long enough, but since clients can create it, maaabye some day someone will use dynamic data in document/collection ID, such as "chatBetweenUser8491568495AndUser879489465". Could this limit be editable in ENV? https://github.com/appwrite/rfc/pull/26/files#diff-f081b69425c96526b8f8e5d6a842fd7e6fcb328c90f0ec2c483d631cb5199543R71
> > * What happens if I try to create function that already exists, or maybe even tag? It will throw error or update it? Update seems to be more friendly solution but may break a lot of current setups.
> 
> @Meldiron nice insight.
> 
> * I should include Functions
> * I believe IDs will be case sensitive
> * I don't think auto completion will be possible at the moment, can't figure out in my head how it would work
> * I think we want to limit the id so that we can validate, keep it secure
> * If id already exists it should throw and error, don't like the idea of updating it, might be friendly but error prone

Agree about updating the ID, it might create lots of confusion, especially around permissions. We might be able to create an internal auto-completion service for the console, that might also be consumed by the VS Code extension, but this requires more research around ID search and the relevant index performance.

018-custom-id-support.md
```diff
@@ -0,0 +1,122 @@
...
+Resources that can have Custom IDs
+- Projects
+- Collections
+- Documents
+- Buckets
+- Files
+- Users
+- Function tags
+- Tasks
+- Webhooks
```
eldadfux, line 63: As more as I think about it, for both consistency and ease of use across the API, any resource should be able to get a custom ID as long as we provide an easy way to auto-generate it.

This includes:
- Teams
- Memberships
- Sessions
- Executions

and more...

Meldiron:
I got a bit confused with these custom IDs, as with this feature, my function (or collection, or document) will ALWAYS have the same ID and name. Is there a use-case when I would use a different custom ID and name?

eldadfux:
> I got a bit confused with these custom IDs, as with this feature, my function (or collection, or document) will ALWAYS have the same ID and name. Is there a use-case when I would use a different custom ID and name?

Not sure I understand what you mean. You can choose a custom ID, you can't update it. You can choose a name and you CAN update it. If you need to update the custom ID, you can just create a new resource.

Meldiron:
> > I got a bit confused with these custom IDs, as with this feature, my function (or collection, or document) will ALWAYS have the same ID and name. Is there a use-case when I would use a different custom ID and name?
> 
> Not sure I understand what you mean. You can choose a custom ID, you can't update it. You can choose a name and you CAN update it. If you need to update the custom ID, you can just create a new resource.

I see, makes sense.

PR #28, opened 2021-07-14
feat(cli-2.0): initial draft

christyjacob4:
preview 
https://github.com/appwrite/rfc/blob/cli-2.0/018-cli-2.0/018-cli-2.0.md


018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
+# Title <!-- What do you want to call your `awesome_feature`? -->
```
eldadfux: Missing title

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+The refactored command will look like this 
+
+```sh
+appwrite functions createTag --functionId=[ID] --command=[COMMAND]
+```
+
+This corresponding appwrite issue will also need to be tackled as we implement this.
+https://github.com/appwrite/appwrite/issues/1316
+
+### 🟢 The `.appwrite.json` File
```
eldadfux: Is there any benefit from a hidden file? Should we just use `appwrite.json`?

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+✅️ FunctionOne
+✅️ FunctionTwo
+⭕️ FunctionThree
+
+# Usage 2
+appwrite deploy functions --all
+# Deploys all the functions specified in .appwrite.json
+
+# Future support for static hosting
+appwrite deploy site
```
eldadfux, line 236: We should choose a consistent name in the future (site/hosting).

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+    "project" : "",
+    "functions": [
+        {
+          "name" : "My awesome function 💪",
+          "path" : "./FunctionOne",
+          "command" : "python main.py",
+          "runtime": "python3.9"
+        },
+    ],
+    "hosting": {}
```
eldadfux:
```suggestion
    "hosting": []
```

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+{
+    "project" : "",
+    "functions": [
+        {
+          "name" : "My awesome function 💪",
+          "path" : "./FunctionOne",
+          "command" : "python main.py",
+          "runtime": "python3.9",
+          // This section is useful when we allow cloud functions to be tested locally
+          "environment_variables" :{
```
eldadfux:
```suggestion
          "vars" :{
```

This is the name used here: https://appwrite.io/docs/server/functions#functionsCreate

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+            "KEY 1" : "VALUE 1",
+          }
+        },
+    ],
+    "hosting": {}
```
eldadfux:
```suggestion
    "hosting": []
```

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,312 @@
...
+  ```sh
+  $ appwrite login
+
+  # Browser based authentication continues. After a successful login, the returned token and user ID are stored in a hidden preferences file.
+  ```
+
+  1. The command first finds an available local port. 
+  2. Construct a redirect URL using the available port `https://localhost:1234`
+  3. Spin up a local http server listening to on the available port ( 1234 in this case )
+  4. Opens the browser with a request to either the local appwrite server or appwrite cloud `/login?callback=https://localhost:1234` with the redirect URL as one of the query parameters.
```
eldadfux:
```suggestion
  4. Opens the browser with a request to either the local appwrite server or appwrite cloud `/authorize/cli?callback=https://localhost:1234` with the redirect URL as one of the query parameters.
```

This will also allow us to use `/authorize/vscode?callback=https://localhost:1234` in the future. We should also list some of the changes that will be required in the console itself. For example:
1. Custom redirect after login - if the user is not logged in the console, he should be redirect to the signin page, and after the signin he should be brought back to `/authorize/cli?callback=https://localhost:1234`.

2. We should also add protection against [open-redirect attack](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html).

eldadfux (changes requested):

018-cli-2.0/018-cli-2.0.md
christyjacob4: done :+1:

018-cli-2.0/018-cli-2.0.md
christyjacob4: I made it hidden under the assumption that we will need to store secrets in there. But since we planned to store the secrets in another file, this can be changed to  `appwrite.json`

018-cli-2.0/018-cli-2.0.md
christyjacob4: Yes sure :+1:

PineappleIOnic (approved):

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,393 @@
...
+
+Here are some example usages 
+```sh
+# Usage 1
+appwrite generate function --name="FunctionOne" --runtime=python-3.9
+# Creates a folder called FunctionOne with a template for a sample python function.
+temp
+├── appwrite.json
+└── FunctionOne
+    ├── .appwrite
```
lohanidamodar, line 256: May be some information regarding `.appwrite` file will be useful as well

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,393 @@
...
+    ├── main.py
+    └── requirements.txt
+
+
+# Usage 2
+appwrite generate function --name="FunctionTwo" --runtime=node-16.0
```
lohanidamodar: May be we can introduce one more parameter here `--template` which should be a link to an accessible github repository/branch so that devs can customize to use their own templates as well as our more advance templates, even our demos can be used as template easily using this feature?

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,393 @@
...
+```json
+{
+    "project" : "",
+    "functions": [
+        {
+          "name" : "My awesome function 💪",
+          "path" : "./FunctionOne",
+          "command" : "python main.py",
+          "runtime": "python3.9",
+          // This section is useful when we allow cloud functions to be tested locally
```
lohanidamodar, line 155: JSON can't have comments 😉

lohanidamodar (commented):
Some suggestions

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+Deploying FunctionOne.. 
+FunctionOne already exists 
+Updating FunctionOne with new tag
+Do you want to activate FunctionOne? [Yes/No]
+
+
+# Usage 2
+appwrite deploy functions --all
+# Deploys all the functions specified in appwrite.json
+
```
abnegate, line 234: What do you think about adding a 4th usage for CI environments to deploy _some_ of the projects functions:

```bash
appwrite deploy functions --names "My awesome function 💪, FunctionTwo" or
appwrite deploy functions --names "My awesome function 💪" "FunctionTwo"
```

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+### 🟢 The `appwrite deploy` command
+
+The `appwrite deploy` command is a convenient wrapper around three main commands
+* Create Function
+* Create Tag
+* Update Tag to activate it 
+
+```sh
+# Usage 1
+appwrite deploy functions --confirm
```
abnegate, line 218: For simplicity of use, one of `--confirm` or `--all` could be a default

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+---
+
+We will introduce an `appwrite init function` command that does the following
+* Generate a function / static site template. 
+* Make an entry in the `appwrite.json` file. Throw an error if the file doesn't exist.
+
+Here are some example usages 
+```sh
+# Usage 1
+appwrite init function --functionId="abc" --name="FunctionOne" --runtime=python-3.9
```
abnegate, line 251: Maybe we could allow `runtime` to be a language and use the latest version available by default? e.g.:

```bash
appwrite init function --functionId="abc" --name="FunctionOne" --runtime=python
```

Which would use `python-3.9` or whichever is newest at the time

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+
+
+## Migration of the CLI from a Server SDK to a Console SDK
+---
+ 
+Until now, the CLI was based off of our Server Side Swagger Spec. While this was a good start, it limits our functionality to a Server SDK. Ideally we want the CLI to be an alternative to the Console SDK ( aka the Appwrite Dashboard ). Here are some changes that we will need to make to adhere to the server spec.
+
+After the implementation of this RFC, the CLI will behave more like the Appwrite Console rather than a Server SDK. This accounts for a new form of Authentication that is similar to the way we handle logins in the Appwrite console. We use a cookie.
+
+We will introduce a new command `appwrite login` that will work in one of either ways.
```
kodumbeats, line 71: Do we need to pick one or the other? I understand that we should think to support OAuth login, but it's not a great CLI experience if a browser is required to login.

018-cli-2.0/018-cli-2.0.md
kodumbeats: This almost feels like an `-i --interactive`mode, like an interactive rebase in git?

018-cli-2.0/018-cli-2.0.md
kodumbeats: Is it possible to have everything in `appwrite.json`? I think it's a bit confusing to have two different types of config files.

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+└── FunctionTwo
+    ├── index.js
+    ├── node_modules
+    └── package.json
+
+
+# Usage 3
+appwrite init function
+
+Give your function an ID
```
kodumbeats, line 279: Is it possible to validate this ID with the Key validator as a part of the CLI? We could provide real-time feedback even before a request is made.

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+This section is intended to encourage you as an author to think about the
+lessons from other software, provide readers of your RFC with a fuller picture.
+If there is no prior art, that is fine - your ideas are interesting to us
+whether they are brand new or if it is an adaptation from other software.
+
+Write your answer below.
+-->
+
+This RFC is inspired by the Firebase CLI.
+
```
kodumbeats, line 412: Does the Fission CLI have any insights to offer?
https://docs.fission.io/docs/usage/function/functions/

018-cli-2.0/018-cli-2.0.md
PineappleIOnic: Maybe most major configs are done using appwrite.json then .appwrite can be used for API keys and project keys as it'll be hidden by default right? Aslong as we remind users not to commit .appwrite they can share their project with ease

Meldiron:
## Local testing

Defining testing environments inside `appwrte.json` looks promising, but will we be able to write multiple tests? I believe all scripts are supposed to return different outputs for different inputs, so we should have an option to write multiple tests. For example

- Register newsletter should return "success" if the email is "matejbaco2000@gmail.com"
- It should return "error" if the email is "h93grnfb9qewn"
- It should return "error" if the Recaptcha result is incorrect

I can see a solution where instead of "vars" we have "tests" objects like this:
```
tests: [
  {
      vars: { email: "matejbaco2000@gmail.com" },
      name: "it should return success with correct email",
      expectedLog: "success"
    },
  {
      vars: { email: "blabla" },
      name: "it should return error with wrong email",
      expectedLog: "error"
    }
]
```

This is a pretty simple example because we might be looking for more statements such as `isNull` or `isGreaterThan`. the `expectedLog` could follow syntax similar to new database refactoring filters so it could look something like:
```
tests: [
  {
      vars: { a: 5, b: 8, method: "+" },
      name: "it should equal  5 + 8 = 13",
      expectedLog: "this.result.equalTo(13)" // for output: { result: 13 }
    },
  {
      vars: { a: 5, b: 8, method: "-" },
      name: "it should equal  5 - 8 = -3",
      expectedLog: "this.result.equalTo(-3)" // for output: { result: -3 }
    },
  {
      vars: { a: 2423423, b: 1231412323, method: "*" },
      name: "it should equal  be huge number",
      expectedLog: "this.result.higherThan(999999999)"
    },
   {
      vars: { a: 5, b: 0, method: "/" },
      name: "it should throw error  5 / 0",
      expectedLog: "this.errorMsg.equalTo('Wrong input')" // for output: { result: null, errorMsg: "Wrong input" }
    },
]
```

## .appwriteignore

If we build the application, we only want to deploy the `build` folder. That would be possible.
Problem is, what if we don't build our code and we need to deploy the whole folder? This folder can include `.git`, `.env` and other files that may include secret information that we don't want to ship with a build.

Also, if we have a build process in place, I may have `node_modules` filled with a bunch of development libraries that are not required for production. Can I somehow remove them from deploy to keep it smaller?

## Database setup

Since we can set up sites and function, can we also add an option to define collection structure using appwrite CLI? There could be `database` folder that would hold migrations and seeds, similar to how MySQL does it ([Knex](https://github.com/knex/knex) for example). Migrations define the scheme of a collection and seeds fill the collection with sample documents. This would also get really related to custom IDs in future, since I probably want to use custom IDs for seed documents (such as Settings document)

Meldiron:
## Function templates

Current RFC includes command:
```
appwrite init function --functionId="abc" --name="FunctionOne" --runtime=python-3.9
```

For python, it generates files `main.py` and `requirements.txt`. For Node, I would expect `package.json`and `app.js`.

What if we want to use Typescript? In the end, we are still going to use Node (or Deno) runtime, but the function structure needs to be way different. First of all, we need `tsconfig.json` and code In `src` folder. Then, we need a `build` function. Finally, we need to deploy a specific folder (+ node_modules?).

In my opinion, the init function could get one more parameter:
```
appwrite init function --functionId="abc" --name="FunctionOne" --runtime=python-3.9 --template=python
```

There will be multiple `templates` defined by Appwrite for the most common use cases. This parameter could also support GIT reference, so people can create their own templates. I don't think there is much that the template needs to know. Information such as relative path, function name or runtime can be easily passed using handlebars.

stnguyen90:
The current CLI is pretty slow because it has to spin up a docker container to execute one command and then tear down.

In addition, passing objects into the CLI is incredibly difficult because of how the data is passed into the container and parsed via the CLI code ([Discord conversation](https://discord.com/channels/564160730845151244/564161373148414012/902679516567199744))

Any plans on rewriting the CLI to not be a docker image like this?

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+          "command" : "python main.py",
+          "runtime": "python3.9",
+          // This section is useful when we allow cloud functions to be tested locally
+          "vars" :{
+            "KEY 1" : "VALUE 1",
+          }
+        },
+    ],
+    "hosting": []
+}
```
wess, line 162: I would like to suggest YAML or even TOML, support for them everywhere, easier to read and less prone to type-o frustration with quotes and such. Also... comments.

018-cli-2.0/018-cli-2.0.md
```diff
@@ -0,0 +1,431 @@
...
+```
+
+There are two scenarios here 
+* A developer would like to link the directory to an existing project
+
+```sh
+appwrite init
+
```
wess, line 170: its this to add the Appwrite stack, to setup for functions or just create the config file with the project id? If project, will it also create the project for us on our endpoints?

018-cli-2.0/018-cli-2.0.md
wess: could also follow the common CLI trends with `-y` instead of a --confirm or --all

018-cli-2.0/018-cli-2.0.md
wess: With this and others, I believe we should name space the commands `functions deploy` (and looking further down, `site deploy -y`

018-cli-2.0/018-cli-2.0.md
wess: Like above: `appwrite function init` -- This helps when printing help screens too, because you can. order them by the namespaces for readable grouping

018-cli-2.0/018-cli-2.0.md
wess: Agreeing with @kodumbeats - If we are going to have a `.appwrite` directory, why not have `config.yaml` in there (example, how .git is setup) or just have `appwrite.yaml`

christyjacob4: @lohanidamodar  `.appwrite` is actually the folder which some runtimes use to store their dependencies like Python and ruby  . Similar to node modules in npm projects

christyjacob4: @kodumbeats `appwrite.json` is a file that is expected to be a part of version control since it will have all the information related to a project like the collections, functions and so on and hence allow them to recreate their project environment . Having secrets in this file doesn't seem like a good idea

stnguyen90:
@christyjacob4, what are the next steps on this?

PR #30, opened 2021-07-28
New Function Execution Model

PineappleIOnic:
This is a draft for the RFC which will form the basis of the new Function Execution Model


018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+The executor will have multiple tasks but the main one is to spin up runtimes when they are needed and to update their code whenever functions are updated. Runtimes will now be web servers based of their relevant languages and when code is added it will be added to the Runtime web server's router.
+
+The executor will also act like a router itself, when API Requests that are for functions are received it will be the executor's job to correlate them to the relevant runtimes and forward them for processing. When the Runtime's web server is done processing the request and returns a response the executor will return the response to whatever client requested it.
+
+#### Scaling
+Scaling this solution will need to be worked on in the RFC but a solution I thought of would be to have the executor to keep track of resources being used by runtimes and the total resources being used hits a certain threshold another instance of the runtime will be spun up and requests will be load balanced onto the multiple runtime instances.
```
lohanidamodar: So the executor will act as both router and load balancer as well?

018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+With this redesign of how cloud functions work the functions worker will no longer have to deal with spinning up docker containers but will instead use the executor server. Which will handle spinning up containers on behalf of the functions worker.
+
+The executor will have multiple tasks but the main one is to spin up runtimes when they are needed and to update their code whenever functions are updated. Runtimes will now be web servers based of their relevant languages and when code is added it will be added to the Runtime web server's router.
+
+The executor will also act like a router itself, when API Requests that are for functions are received it will be the executor's job to correlate them to the relevant runtimes and forward them for processing. When the Runtime's web server is done processing the request and returns a response the executor will return the response to whatever client requested it.
```
lohanidamodar: I think here we need more technical description of how the executor it self will look, may be API signatures,
how worker/executor will communicate, some sample structures etc?

018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+**What is the context or background in which this problem exists?**
+
+Cloud Functions are currently a bit limited, this new design idea will allow for more customizability and control of how functions will execute and run for users.
+
+**Once the proposal is implemented, how will the system change?**
+
+- A new process will be introduced to execute functions
+- All current runtimes will be updated to use a web server as it's core
+- The Functions worker will be stripped of all direct docker interactions and will instead call the executor
```
lohanidamodar: May be we can also talk about how performance will be impacted after this change?

lohanidamodar (commented):
Some queries and feedback

018-function-execution-model-refactor.md
abnegate: I agree here

018-function-execution-model-refactor.md
kodumbeats: Looks like Fission's executor has two execution types, depending on if the serverless function needs to scale due to load. I think this was their solution to avoid having the executor also act as load balancer.
ref: https://docs.fission.io/docs/architecture/executor/#new-deployment

018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+With this redesign of how cloud functions work the functions worker will no longer have to deal with spinning up docker containers but will instead use the executor server. Which will handle spinning up containers on behalf of the functions worker.
+
+The executor will have multiple tasks but the main one is to spin up runtimes when they are needed and to update their code whenever functions are updated. Runtimes will now be web servers based of their relevant languages and when code is added it will be added to the Runtime web server's router.
```
kodumbeats: Should the executor also be responsible for cleaning up function containers when they're no longer necessary?

018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,89 @@
...
+
+**What is the context or background in which this problem exists?**
+
+Cloud Functions are currently a bit limited, this new design idea will allow for more customizability and control of how functions will execute and run for users.
+
+**Once the proposal is implemented, how will the system change?**
+
+- A new process will be introduced to execute functions
+- All current runtimes will be updated to use a web server as it's core
+- The Functions worker will be stripped of all direct docker interactions and will instead call the executor
```
kodumbeats: If the Functions worker is stripped of all its direct Docker work, what will be left?

018-function-execution-model-refactor.md
eldadfux: I think it should, but with a simple timer that removes all runtimes that didn't get a call every X minutes. The check can also run in an interval for every X minutes.

018-function-execution-model-refactor.md
eldadfux: Scheduled and async executions?

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+It will act like so:
+
+1. Check if the runtime is running, if not spin up the runtime.
+2. Make a request to the runtime passing the arguments recieve through the API request.
+3. If the request is asyncronous then return the API request now with a execution ID like it does now.
+4. When we receive a response back from the runtime we will do 1 of two things
+    If the request is synchronous:
+        - Wait for the response from the runtime and return it as the response to the API request.
+    If the request is asynchronous:
+        - Store the result for later use through the Get Execution API endpoint.
```
kodumbeats: Where is a function defined as synchronous or asynchronous? Would this be an extra param on `POST /v1/functions`?

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+
+It will act like so:
+1. Extract the code tarball into the enviroment.
+2. Check if the server is running, if it is then stop the server.
+3. also check if the runtime exists, if it doesn't then create the runtime.
+4. Map the new code into the runtime environment using a volume
+5. Finally bring the runtime back up and start the server again.
+
+#### Watchdog
+Watchdog will be apart of the executor running in the background dealing with all the running runtimes and monitoring their health.
```
kodumbeats: So how many processes will be running in the Executor container? Best Docker practice is to limit one process per container.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+Every X minutes (defaults will need to be determined) watchdog will call a health endpoint which will be built into all enviroment web servers.
+If the web server returns a '200' then the runtime is health and watchdog will continue onto the next runtime however if the runtime does not respond in a set time or responds incorrectly then the runtime will be restarted and watchdog will continue monitoring it.
+If after a certain amount of restarts the runtime still does not respond correctly then watchdog will log a error and inform the user that something is wrong
```
kodumbeats: How does this watchdog error propagate back to the user? Will we need additional endpoints for `v1/functions/:functionId/executions` for this?

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+https://docs.fission.io/docs/architecture/
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+ - Scaling this solution even more
+ - Using things such as [Amazon Firecracker](https://firecracker-microvm.github.io/)
+ - Will the executor handle load balencing itself or will we use another service to deal with it?
+ - More technical indepth discussion is needed for figuring out API's
```
kodumbeats: Why are we leaving this as an unresolved question? This is pretty core to the RFC overall.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+
+The executor will have multiple tasks but the main one is to spin up runtimes when they are needed and to update their code whenever functions are updated. Runtimes will now be web servers based of their relevant languages and when code is added it will be added to the Runtime web server's router.
+
+The executor will also act like a router itself, when API Requests that are for functions are received it will be the executor's job to correlate them to the relevant runtimes and forward them for processing. When the Runtime's web server is done processing the request and returns a response the executor will return the response to whatever client requested it.
+
+#### Scaling
+Scaling this solution will need to be worked on in the RFC but a solution I thought of would be to have the executor to keep track of resources being used by runtimes and the total resources being used hits a certain threshold another instance of the runtime will be spun up and requests will be load balanced onto the multiple runtime instances.
```
kodumbeats: Will the executor store any of this data, or will we just be querying the Docker API for stats? Will we want to monitor with another solution?
ref: https://docs.docker.com/engine/api/v1.41/#operation/ContainerStats

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+
+- Its interaction with other parts of the system is clear
+- It is reasonably clear how the contribution would be implemented
+- Dependencies on libraries, tools, projects or work that isn't yet complete
+- New API routes that need to be created or modifications to the existing routes (if needed)
+- Any breaking changes and ways in which we can ensure backward compatibility.
+- Use Cases
+- Goals
+- Deliverables
+- Changes to documentation
```
kodumbeats, line 95: Can we outline the necessary documentation we'll want to release alongside this feature?

kodumbeats (changes requested):

018-function-execution-model-refactor.md
kodumbeats: Looking at the Architecture diagram, there's no significant workflow difference between sync and async functions - in fact, the Functions worker doesn't appear to have any function at all if we follow that document.

With so few responsibilities, I think it's confusing to have both a Functions worker (responsible for very little) alongside the Executor, which does basically everything else. Maybe there's a more appropriate name for the Functions worker after this change?

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+If the web server returns a '200' then the runtime is health and watchdog will continue onto the next runtime however if the runtime does not respond in a set time or responds incorrectly then the runtime will be restarted and watchdog will continue monitoring it.
+If after a certain amount of restarts the runtime still does not respond correctly then watchdog will log a error and inform the user that something is wrong
+with the code running in the runtime. It will also stop attempting to restart it to save on resources.
+
+Watchdog will also deal with cleaning up runtimes that haven't recieve a request in a set amount of time in order to save on resources, aswell as removing
+runtimes that are no longer needed.
+
+#### Flowchart Visualisation:
+
+![Flowchart Visualisation](flowchart.png)
```
kodumbeats: Why does this flowchart not include how the Functions worker fits into the new execution model?

018-function-execution-model-refactor/018-function-execution-model-refactor.md
eldadfux: Agree with @kodumbeats. This is a big overhead and probably not the best pattern for a Docker container.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+
```
eldadfux, line 83: image
This part is also not correct. We don't necessarily need to start the runtime when a tag is activated. It's easier to delegate this responsibility to the first execution. We also don't need to stop the previous runtime. It can be stopped automatically after inactivity by the executor.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
PineappleIOnic: Watchdog could be split away into it's own process if that is the case :+1:

018-function-execution-model-refactor/018-function-execution-model-refactor.md
eldadfux: I would try to avoid dealing with the scaling problem at this layer, although it might have some advantages. I think this can be part of a completely separate PR/RFC for autoscaling Cloud Functions or other Appwrite components.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
eldadfux: The only thing this endpoint should do is either execute the function directly using the executor (sync http request), or send the execution to the workers queue (async - similar to the current implementation).

No need for this endpoint to know if runtime is up or not, the executor will make sure the execution is successful from both the Appwrite API or worker.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+`PATCH /v1/functions/{id}/tag` - Update function tag API endpoint
+The tag update endpoint will also recieve a small rewrite for the request moving it into the executor.
+
+It will act like so:
+1. Extract the code tarball into the enviroment.
+2. Check if the server is running, if it is then stop the server.
+3. also check if the runtime exists, if it doesn't then create the runtime.
+4. Map the new code into the runtime environment using a volume
+5. Finally bring the runtime back up and start the server again.
```
eldadfux: Again the Appwrite API shouldn't care if a runtime is up or no. Actually this endpoint shouldn't change at all at the scope of the RFC.

The executor is the only component interacting and responsible for starting a runtime, checking its status or bringing it down.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Watchdog
```
eldadfux: This is out of the context of this RFC. The usage of watchdog as a component in Appwrite will create a tight coupling between the application and monitoring layer + docker can potentially handle this failures. This is also not clear how watchdog is suppose to play well in the bigger picture of our Architecture. Implementing something like watchdog is not something I think we should discuss as part of this RFC.

eldadfux (changes requested):

018-function-execution-model-refactor/018-function-execution-model-refactor.md
PineappleIOnic: Yes, this will be a extra parameter on `POST /v1/functions/{id}/execute` I'll update the better RFC to reflect this

018-function-execution-model-refactor/018-function-execution-model-refactor.md
PineappleIOnic: I've removed watchdog from the RFC for now, so we can better focus on the core of the implementation

018-function-execution-model-refactor/018-function-execution-model-refactor.md
PineappleIOnic: I have removed the Scaling section for now

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,111 @@
...
+The executor will also act like a router itself, when API Requests that are for functions are received it will be the executor's job to correlate them to the relevant runtimes and forward them for processing. When the Runtime's web server is done processing the request and returns a response the executor will return the response to whatever client requested it.
+
+The executor will be built using a Swoole HTTP Server and will be stored in the file: `app/executor.php`. The executor will need to be on the same network as appwrite so the two can communicate.
+
+#### API Endpoints
+This rewrite will add functionality to the API endpoints that are currently used by the functions worker. The executor will take over these endpoints while the rest will stay underneath the functions worker.
+
+`POST /v1/functions/{id}/executions/` - Execute function API endpoint
+
+The standard execution endpoint for functions will have functionality added for syncronous executions aswell as a new parameter in the JSON body to determine if this execution will be syncronous or asyncronous, this parameter will be called `isSync` and will be a boolean.
```
kodumbeats: I'm not sure that this param should be on the `POST /executions` route, but rather the `POST /tags` endpoint. Since sync and async serverless functions will be written differently (async scripts write to console log, sync scripts handle requests&responses), their execution models can't be interchanged. I think that we should store this fact in the same data structure as the rest of the tag info.

018-function-execution-model-refactor/018-function-execution-model-refactor.md
```diff
@@ -0,0 +1,111 @@
...
+If it's syncronous then it will execute the functions directly using the executor and return the response, returning
+a error with a non 200 http code if anything goes wrong.
+
+`PATCH /v1/functions/{id}/tag` - Update function tag API endpoint
+The tag update endpoint will also recieve a small rewrite for the request moving it into the executor.
+
+It will act like so:
+1. Extract the code tarball into the enviroment.
+2. also check if the runtime exists, if it doesn't then create the runtime.
+3. Map the new code into the runtime environment using a volume
```
kodumbeats: This work happens on first execution for async functions as of now. Let's keep this behavior for sync functions as well?

kodumbeats (changes requested):
Let's get a new architecture diagram in place for the executor, as it contains many of the implementation details we want for this RFC :)

018-function-execution-model-refactor/018-function-execution-model-refactor.md
PineappleIOnic: Alright, that sounds like a good idea

PR #31, opened 2021-08-10
Phone Auth RFC

TorstenDittmann:
https://github.com/appwrite/rfc/blob/rfc-phone-auth/020-phone-auth.md

### DIFF
--- a/020-phone-auth.md
+++ b/020-phone-auth.md
@@ -212,4 +212,4 @@ This could potentially be solved with the already existing authentication method
 
 <!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
 
-This implementation will allow more implementation in the future and also decouple the user account from an E-Mail address.
+This implementation will allow more providers implementation in the future and also decouple the user account from an E-Mail address.
--- a/020-phone-auth.md
+++ b/020-phone-auth.md
@@ -1,0 +1,96 @@
+# Title <!-- What do you want to call your `awesome_feature`? -->
+
+- Implementation Owner: (your github @username)
+- Start Date: (today's date, dd-mm-yyyy)
+- Target Date: (expected date of completion, dd-mm-yyyy)
+- Appwrite Issue:
+  [Is this RFC inspired by an issue in appwrite](https://github.com/appwrite/appwrite/issues/)
+
+## Summary
+
+[summary]: #summary
+
+<!-- Brief explanation of the proposed contribution. Write your answer below. -->
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+<!-- Write your answer below. -->
+
+**What is the context or background in which this problem exists?**
+
+<!-- Write your answer below. -->
+
+**Once the proposal is implemented, how will the system change?**
+
+<!-- Write your answer below. -->
+
+<!-- Please avoid discussing your proposed solution. -->
+
+## Design proposal (Step 2)
+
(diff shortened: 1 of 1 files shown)


020-phone-auth.md
```diff
@@ -0,0 +1,215 @@
...
+│      End       │
+│                │
+└────────────────┘
+```
+
+### Providers Endpoints
+
+This addition also needs to have more endpoints, which are needed to see which providers are linked to which account and also delete a provider from a user.
+
+#### `GET: /v1/account/providers`
```
lohanidamodar, line 115: Why not simply return providers with user object during account.get?
1. That way we will only need a delete endpoint for the provider in Account and Users service,
2. Account service can add provider during session create if a session already exists
3. we only need add provider endpoint for Users service

020-phone-auth.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Completion
+
+This step will compare the token's `$id` and `secret` created in the initialization step. 
+#### `POST: /v1/account/sessions/phone/complete`
+
+The payload must contain the `$id` of the Token and the 6 digit number send to the phone representing the Token's `secret`. If both values are valid, this endpoint will either login an existing user or create a new one if the phone number isn't present on any of the user provider attribute.
+
+### Flow
+```
```
lohanidamodar, line 74: I like the Ascii flowchart,
https://asciiflow.com/

020-phone-auth.md
```diff
@@ -0,0 +1,215 @@
...
+
+This could potentially be solved with the already existing authentication methods, but when a user is already logged in - will append the used provider to the current user.
+
+### Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+<!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
+
+This implementation will allow more implementation in the future and also decouple the user account from an E-Mail address.
```
lohanidamodar:
```suggestion
This implementation will allow more providers implementation in the future and also decouple the user account from an E-Mail address.
```

020-phone-auth.md
```diff
@@ -0,0 +1,215 @@
...
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+#### Are the required credentials consitent across multiple Third Party Providers?
+
+#### How de we add new providers to a user?
+
+This could potentially be solved with the already existing authentication methods, but when a user is already logged in - will append the used provider to the current user.
```
lohanidamodar, line 207: 1. If request is made from account endpoint and there is an active session, add the provider to the current user
2. For users service, there should be separate endpoint to add the provider

Is there something I don't understand here?

020-phone-auth.md
```diff
@@ -0,0 +1,215 @@
...
+Write your answer below.
+-->
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+#### Are the required credentials consitent across multiple Third Party Providers?
```
lohanidamodar, line 203: We should research among most popular providers at the moment, should make the decision simpler

lohanidamodar (changes requested):
Small changes, and few questions. 
Rest looks good to me.

020-phone-auth.md
TorstenDittmann: True, should be part of the user object anyway and we can skip this endpoint.

PR #49, opened 2022-02-06
2-facor authentication

Meldiron:
When writing this, I noticed we currently don't really protect endpoints with passwords. For instance, when deleting an account, the user should provide a password. Appwrite issue has been created on this topic, and it's not the scope of this RFC.

Appwrite issue: https://github.com/appwrite/appwrite/issues/2752

RFC preview: https://github.com/appwrite/rfc/blob/feat-2fa/019-2fa-support.md

### DIFF
--- a/019-2fa-support.md
+++ b/019-2fa-support.md
@@ -1,0 +1,134 @@
+# 2-factor authentication
+
+- Implementation Owner: @Meldiron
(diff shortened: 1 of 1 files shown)


Meldiron:
Appwrite already has 2FA

PR #57, opened 2023-05-16
Create RFC for Identities

stnguyen90:
Rendered: https://github.com/appwrite/rfc/blob/feat-023-connected-accounts/023-identities/README.md

## What does this PR do?

Proposal for separating OAuth2 from Sessions

## Related PRs and Issues

TBD

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

### DIFF
--- a/023-identities/README.md
+++ b/023-identities/README.md
@@ -1,0 +1,335 @@
+# Identities <!-- What do you want to call your `awesome_feature`? -->
+
+- Implementation Owner: @stnguyne90
+- Start Date: 12-06-2023
+- Target Date: 01-07-2023
+- Appwrite Issue:
+  * https://github.com/appwrite/appwrite/issues/3924
+  * https://github.com/appwrite/appwrite/issues/3680
+  * https://github.com/appwrite/appwrite/issues/1192
+
+## Summary
+
+[summary]: #summary
+
+<!-- Brief explanation of the proposed contribution. Write your answer below. -->
+
+Allow users to connect multiple accounts to their Appwrite account regardless of the email address in the external system.
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+<!--
+What problem are you trying to solve? Explain the context or background in which this problem exists.
+Please avoid discussing your proposed solution.
+-->
+
+At the moment, external accounts from an OAuth2 provider are connected to Appwrite via email. When a user creates an OAuth2 session, the email provided by the OAuth2 provider is looked up in Appwrite to find the the existing related user. If the user is not found, a new user is created. The problem with this is a person may have different emails in different OAuth2 providers which would result in a new user being created.
+
+Another problem with our current design is the OAuth2 information is integrated into the session. Because of that, once a session is deleted, there is no information about the OAuth2 provider that was used to create the session (and possibly user).
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+<!--
+This is the technical portion of the RFC. Explain the design in sufficient detail, keeping in mind the following:
+
+- Its interaction with other parts of the system is clear
+- It is reasonably clear how the contribution would be implemented
+- Dependencies on libraries, tools, projects, or work that isn't yet complete
+- New API routes that need to be created or modifications to the existing routes (if needed)
+- Any breaking changes and ways in which we can ensure backward compatibility.
+- Use Cases
+- Goals
+- Deliverables
+- Changes to documentation
+- Ways to scale the solution
+
+Ensure that you include examples and code snippets to allow the community to understand the proposed solution. **It would be best if the examples use naming conventions that you intend to use during the actual implementation to suggest changes early on during the development.**
+
+Write your answer below.
+
+-->
+
+A user will be able to have multiple identities where each identity is a connection to an OAuth2 provider.
+
+A user should be able to:
+
+1. Connect their account to an OAuth2 provider, creating an identity
+2. List their identities
+3. Update their identity
+   * This will refresh the access token
+4. Disconnect their account from an OAuth2 provider, deleting the identity
+5. Create an OAuth2 session with an existing account without a connected OAuth2 provider
+   * This will automatically create an identity
+6. Create an OAuth2 session without creating an account (backwards compatibility)
+
+A developer should be able to:
+
+1. List a user's identities
+2. Delete a user's identity
+
+### API Endpoints
+
+<!--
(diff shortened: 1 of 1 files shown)


023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+##### Create Connected Account
+
+The existing Create OAuth2 Session endpoints will be re-used to create a connected account
+
+**GET /v1/account/sessions/oauth2/{provider}**
+
+##### List Connected Accounts
+
+**GET /v1/account/connected**
```
eldadfux: Connected accounts is very misleading. When I first heard of it I thought it was about connecting 2 existing Appwrite accounts and allowing to switch between them. This features is actually just refactoring of how we handle OAuth2.

eldadfux:
```suggestion
**GET /v1/account/oauth2/connections**
```

Connected accounts is very misleading. When I first heard of it I thought it was about connecting 2 existing Appwrite accounts and allowing to switch between them. This features is actually just refactoring of how we handle OAuth2.  Also, the `connected` is a past tense verb where a REST API should be represented by nouns.

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+* queries: it should be possible to filter the queries by provider
+
+Response:
+
+List of [Connected Accounts](#connected-account)
+
+##### Get a Connected Account
+
+**GET /v1/account/connected/{id}**
```
eldadfux:
```suggestion
**GET /v1/account/oauth/connections/{id}**
```

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+* accessTokenExpiry: the date when the access token expires in ISO 8601 format
+* providerRefreshToken: refresh token from the provider
+
+### Flow
+
+#### User
+
+##### Connecting to an OAuth2 provider
+
+1. User makes authenticated API call to `GET /v1/account/sessions/oauth2/{provider}`
```
eldadfux: Is it safe to assume that the main difference here from the existing flow is that if a user is already logged in with Appwrite when creating an OAuth session the OAuth connection will be attached to their current account regardless of the email?

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+The SDKs will need to be updated to support the new endpoints.
+
+### Breaking Changes
+
+<!--
+Do we break any API or SDK backward compatibility?
+If possible, explain what actions we can take to avoid that.
+-->
+
+The existing OAuth2 APIs will still work as they do now. The only difference is a connected account will be created if one does not exist.
```
eldadfux: This is a better place in the RFC to discuss data migrations due to changes in data structure.

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+#### Tests (UI, Unit, E2E)
+
+<!-- 
+Explain how we will test the new feature. 
+You can use "N/A" if this section is not relevant to your proposal.
+-->
+
+### Documentation & Content
+
+<!--
```
eldadfux: Missing some info here.

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+<!-- What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+
+<!-- Write your answer below. -->
+
+1. How to connect an account on mobile?
```
eldadfux: What is the main challenge? Passing the current use session token to the webview?

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+2. Should we support a way to get all connected accounts from a particular provider?
```
eldadfux: We can do this using the list endpoint queries.

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+3. What to do after a connected accounts refresh token no longer works?
```
eldadfux: I would expect maybe something 401, which can notify the developer they need to re-authenticate the OAuth session to obtain a new access token. How often does this happen today for the different providers? how long would an average access token last before expiration?

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+4. What should happen to related sessions when a connected account is deleted?
```
eldadfux: I would leave this unattached. If dev wish to take action here, they can delete relevant sessions with existing APIs. I wouldn't be opinionated and decide that deleting an account is necessarily connected with a security concern, this is not always the case.

023-connected-accounts/README.md
```diff
@@ -0,0 +1,303 @@
...
+5. What should happen to a connected account when the related session is deleted?
```
eldadfux: Nothing. You can logout, it doesn't imply you want your connection removed. If developers want for any reason to tie the two together they can do it using the two relevant APIs.

023-connected-accounts/README.md
eldadfux: If yes, we could consider another form of short living token with limited access rights. We have similar problems with file previews and other resources as well.

eldadfux (changes requested):
Steven, great work, super interesting feature. I left a few comments, thoughts, and questions.

023-connected-accounts/README.md
stnguyen90: Wouldn't it be useful to make this flexible so that we could support other connections? Maybe multiple phone numbers or linking multiple emails? Merging Appwrite users can also be helpful.

023-connected-accounts/README.md
stnguyen90: Yes, during the /redirect.

023-connected-accounts/README.md
stnguyen90: > I would expect maybe something 401

You mean if the user is using an OAuth2 session? This means we would need to check the refresh token on each request or something.

In addition, we may need to check the tokens on each list connected accounts API call so that we properly return the correct status.

023-connected-accounts/README.md
stnguyen90: Updated

023-connected-accounts/README.md
stnguyen90: Added a line.

PR #59, opened 2023-08-31
Server Side Rendering

loks0n:


021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+SSR is popular for a number of reasons:
+
+- It enables search engines to crawl the site and index the content
+- It improves performance by reducing the time to first render, and reducing the amount of JavaScript required to render the page
+- It improves accessibility by ensuring that the page is usable even if JavaScript is disabled
+- It prevents flash of unauthenticated content.
+
+**Problem #1: Accessing basic sessions**
+
+To authenticate using anonymous sessions, email-password sessions and magic URL sessions, the Appwrite SDKs make a POST request to the relevant endpoint. The response body contains the session ID, but not the
```
Meldiron: but not the ...

(probably secret? based on context)

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+
+const account = new Account(client);
+const currentUser = account.get();
+```
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+**Problem #1: Solution A - Return session token in response body**
```
Meldiron: I like solution A, can't see why not.
Such token should be hidden by default, and only exposed when creating session.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+7. Appwrite API sets the session cookie on the Appwrite domain.
+8. Appwrite API redirects the browser back to the client application. **Now, the redirect URL includes a userId & temporary token.** e.g. `myssrapp.com/oauth2/success?userId=loks0n&token=TEMPORARY_TOKEN`
+9. The SSR application makes a POST request to the Appwrite API to exchange the temporary token for a session token.
+
+The SSR application must set up the success page to call the exchange endpoint. The page can then set session cookie on the SSR domain.
+`
+
+**Problem #3: Solution A - setSession helper method**
+
+A new SDK helper method, called `setSession`, to set the session token of future requests.
```
Meldiron: SetSession or SetToken? 🤔

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+![SSR OAuth2 Flow Sequence Diagram](ssr-oauth-flow.png)
+
+1. Browser makes a GET request to the SSR application.
+2. SSR application makes a GET request to the Appwrite API.
+3. Appwrite API redirects the SSR application to the OAuth2 provider.
+4. The SSR application redirects the client to the OAuth2 provider.
+5. User authenticates with the OAuth2 provider.
+6. OAuth2 provider redirects the browser back to the Appwrite API.
+7. Appwrite API sets the session cookie on the Appwrite domain.
+8. Appwrite API redirects the browser back to the client application. **Now, the redirect URL includes a userId & temporary token.** e.g. `myssrapp.com/oauth2/success?userId=loks0n&token=TEMPORARY_TOKEN`
```
Meldiron: Currently OAuth works with 1 line of code.
Wit this, it wont. You will need to add extra logic on page load that looks for those params, and then runs method to finish sign-in process.

Similiar to Magic URL - After sending mail, you need 1 extra page that takes URL parals and finishes the logic.

That's kinda annoying :/ Can we find a way to automate this, somehow? Or simplify?

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+
+**Problem #1**
+
+1.1. Which solution should we implement? A or B?
+
+**Problem #2**
+
+2.1. Should we remove the undocumented workaround?
+
+2.2. Do we need to add a new endpoint to exchange the temporary token for a session token? Can we reuse the existing magic URL exchange endpoint?
```
Meldiron, line 256: I would not reuse.

Buut I like the idea. But it's of a bigger scope.

My idea is that there should be universal "token -> session" endpoint, where token would also know type of session, like `magicUrl` or `oauth2`. Why? Well because then we can add endpoint `users.createCustomSession(userId)`, which returns such token. Then we can exchange this token for session.

This will allow fully custom auth flows. For example, with APpwrite Functions, this will allow logging into account by entering code that Discord bot sent you as private message.. Whatever you want.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+
+2.3. The Create OAuth2 Session endpoints are not included with Server Side SDKs. Should we include them? If so, server side apps need the Location header to forward to the user. How should we make this accessible from the SDK?
+
+**Problem #3**
+
+3.1. Should the helper method be available in Server Side SDKs? How do we seperate requests that are authenticated with a session token, and requests that are authenticated with an API key?
```
Meldiron: I would add them to server sdks, yes. I think they could be useful.
To differentiate we can just use another header name, like `x-appwtite-token`

021-server-side-rendering/021-server-side-rendering.md
loks0n: fixed

021-server-side-rendering/021-server-side-rendering.md
loks0n: What do you mean by `such token should be hidden by default`?

021-server-side-rendering/021-server-side-rendering.md
loks0n: Unfortunately, I can't see a way to simplify this flow: we are at the mercy of browser cookie security.

Here's why:
- We need a place to store a session on the client, and send it with browser GET requests.
- Local storage doesn't work because browsers don't send it with requests. We need to use secure cookies.
- Browsers only send cookies that belong to requested domain.
- Domains may only set cookies for their own domain.

The extra page isn't so much of an issue if we have a universal token -> session endpoint, right?

021-server-side-rendering/021-server-side-rendering.md
loks0n: I feel we generally use token for temporary keys.
This helper could take either the entire session object, or just the secret, in which case `setSecret` or `setSessionSecret` might be worth considering

021-server-side-rendering/021-server-side-rendering.md
loks0n: +1 for Universal 'token -> session' endpoint

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+
+Appwrite Auth is currently tailored towards client side rendered (CSR) applications. This RFC proposes to add support for server side rendered (SSR) applications.
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+**What problem are you trying to solve?**
+
+Many popular web frameworks, such as Next.js, Nuxt.jsm and SvelteKit, support and recommend SSR, by default. When developers attempt to use Appwrite Auth with these frameworks, they run into issues with the current implementation.
```
TGlide:
```suggestion
Many popular web frameworks, such as Next.js, Nuxt.js and SvelteKit, support and recommend SSR, by default. When developers attempt to use Appwrite Auth with these frameworks, they run into issues with the current implementation.
```
Sorry 😬 

Also, I'd say the problem is not only Appwrite Auth. I've also run into issues with file uploads. I've managed to get it working, but it was [troublesome to say the least](https://github.com/TGlide/pluck-svelte/blob/main/src/lib/appwrite/storage.server.ts).

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+import { Client, Account } from "appwrite";
+
+const client = new Client();
+client.setEndpoint("https://cloud.appwrite.io/v1");
+client.setProject("PROJECT_ID");
+
+const account = new Account(client);
+const session = await account.createAnonymousSession();
+const sessionToken = session.token;
+```
```
TGlide: This works, but what do we then do with the sessionToken? How would a more complete flow look like?

For reference, [this is what I do now](https://github.com/TGlide/pluck-svelte/blob/main/src/lib/appwrite/account.server.ts) to achieve SSR in auth.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+
+**Problem #1: Solution B - Include headers within the response**
+
+Return the session cookie in the response headers.
+
+```js
+import { parse } from "set-cookie-parser";
```
TGlide: I don't think we should require an external lib to be able to use cookies. An inbuilt helper would be much friendlier

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+#### What **docs** would support this feature?
+
+- SDK docs and examples for the new helper method.
+- SDK docs and examples for the new token exchange endpoint.
+
+#### What **tutorials** (text/video) might help developers understand this feature scope, capabilities, and possible use-cases?
+
+- Generic Tutorial for using Appwrite for SSR
+- Tutorial for using Appwrite and Next.js for SSR
+- Tutorial for using Appwrite and SvelteKit for SSR
```
TGlide: We should include more frameworks here

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+
+1.1. Which solution should we implement? A or B?
+
+**Problem #2**
+
+2.1. Should we remove the undocumented workaround?
+
+2.2. Do we need to add a new endpoint to exchange the temporary token for a session token? Can we reuse the existing magic URL exchange endpoint?
+
+2.3. The Create OAuth2 Session endpoints are not included with Server Side SDKs. Should we include them? If so, server side apps need the Location header to forward to the user. How should we make this accessible from the SDK?
```
TGlide: By server-side SDK, do you mean `node-appwrite`?

TGlide: If so, would you require to have both `node-appwrite` and `appwrite` installed in an SSR framework? When would each be used?

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+
+**Problem #3**
+
+3.1. Should the helper method be available in Server Side SDKs? How do we seperate requests that are authenticated with a session token, and requests that are authenticated with an API key?
+
+**General**
+
+4.1. Incorporating some of these SSR changes reduces the distinction between the Server Side and Client Side SDKs. Which SDKs would be recommended for SSR applications?
```
TGlide: Exactly, it gets harder to differentiate. Also, I worry about rate-limiting, which would currently happen using SSR, right?

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,296 @@
...
+  value: session.token,
+  path: "/",
+  sameSite: "none",
+  secure: true,
+  httpOnly: true,
+  maxAge: session.expire,
+});
+```
+
+We can provide a helper method for popular SSR frameworks to set the cookie in the response.
```
TGlide: Love this! Would be great for adoption.

021-server-side-rendering/021-server-side-rendering.md
loks0n: >By server-side SDK, do you mean node-appwrite?

Yeah, node-appwrite and the other server SDKs

> If so, would you require to have both node-appwrite and appwrite installed in an SSR framework? When would each be used?

This is what I'm getting at, in an ideal world I think we should have a single SDK that handles any compatibility issues.
Second best and easier option is to recommend using `node-appwrite` in server code and `appwrite` in all client code.

021-server-side-rendering/021-server-side-rendering.md
Meldiron: We have `setKey` for setting API key. So to follow that we should do `setSecret` for setting session secret.

But that sounds so undescriptive.. I would vote for `setSessionSecret`, but then we should consider doing `setApiKey`.

021-server-side-rendering/021-server-side-rendering.md
Meldiron: You had 2 solutions. One to have it in response body, one to set it as response header.

I like setting it in response body. But if in future you do `getSession`, you should no longer get secret. You should only get it once when you first create it.

021-server-side-rendering/021-server-side-rendering.md
Meldiron: I see. I agree it's better to ask developer to write more code than to have it not work on some setups like Safari or phone devices.

Might be worth checking competitors and see if they have setup this complex. if not, maybe they found some trick.

021-server-side-rendering/021-server-side-rendering.md
Meldiron: @eldadfux Do you think it would be worth as part of SSR rework? It is pretty highly requested feature. I think this will be most common blocker for any bigger company.

eldadfux: We can.

021-server-side-rendering/021-server-side-rendering.md
eldadfux: +1

021-server-side-rendering/021-server-side-rendering.md
eldadfux: +1

021-server-side-rendering/021-server-side-rendering.md
eldadfux: I guess for all fullstack / web work you should use `appwrite` for any admin related tasks you should use `node-appwrite` which is basically and admin (or console) like SDK.

021-server-side-rendering/021-server-side-rendering.md
eldadfux: @loks0n let's check where can rate limit get in the way. If our key is email+ip we should be fine. If the abuse key for an endpoint is only for IP it will be useless.

021-server-side-rendering/021-server-side-rendering.md
eldadfux: Yeh, this is a must. Can we include it in the main SDK or do we have to create separated packages for it?

eldadfux:
Can we try and redirect OAuth flow back to the SSR server itself with relevant params (including a secret) on a dedicated route and use it to set the local cookie in a similar way to what we do for a regular session?

021-server-side-rendering/021-server-side-rendering.md
loks0n: Can be in the main SDK :)

021-server-side-rendering/021-server-side-rendering.md
loks0n: Rate limiting is an issue, I've added a section outlining some possible solutions

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,304 @@
...
+const cookieList = cookies();
+cookieList.set(createSessionCookie(session, "my_cookie_name"));
+```
+
+Popular JS frameworks worth including:
+
+- Next
+- SvelteKit
+- Nuxt
+- Remix
```
TGlide, line 354: I'd add Astro to the list

loks0n:
> Can we try and redirect OAuth flow back to the SSR server itself with relevant params (including a secret) on a dedicated route and use it to set the local cookie in a similar way to what we do for a regular session?

Are you suggesting to replace the redirect from the OAuth2 provider to Appwrite with OAuth2 to SSR? The Appwrite session is not yet created at this point, and therefore the server would still need to exchange the temporary token

021-server-side-rendering/021-server-side-rendering.md
loks0n: I could live with `setSecret`, although one advantage of `setSession` is it makes more sense if used in combination with `setKey` as may be required to solve rate limiting

021-server-side-rendering/021-server-side-rendering.md
loks0n: Supabase and Firebase are using the same method, unfortunately.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,398 @@
...
+- https://firebase.nuxtjs.org/tutorials/ssr.html
+- https://github.com/gladly-team/next-firebase-auth
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+**Problem #2**
+
+2.1. Should we remove the undocumented OAuth2 SSR workaround?
```
gewenyu99: Yes. Undocumented work arounds are distracting at best and polluting/attract abuse at worse

021-server-side-rendering/021-server-side-rendering.md
gewenyu99: Heard of requests for this in various avenues. Is a build/scale stage issue.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,398 @@
...
+
+2.2. Do we need to add a new endpoint to exchange the temporary token for a session token? Can we reuse the existing magic URL exchange endpoint?
+
+**General**
+
+4.1. Some of these changes reduce the distinction between the Server and Client Side SDKs. Which SDKs would be recommended for SSR applications?
```
gewenyu99, line 260: Define something that we can clearly explain and suggest in docs. 

If it feels complex to explain, probably too complicated.

021-server-side-rendering/021-server-side-rendering.md
gewenyu99: @loks0n What's the expected effect on package size, build time, etc.

If it's trivial we can add, but I wonder when we draw the line for framework specific baggage in the main SDK

TGlide: @gewenyu99 in general nowadays tree-shaking should deal with most issues regarding bundle size on your app's final output, build time too. The package size would be bigger on the `node-modules` folder though.

Alternatively, you can monorepo the SDK, and add adapters. E.g. `@appwrite.io/adapter-svelte`

021-server-side-rendering/021-server-side-rendering.md
loks0n: If you want to make privileged actions, such as provisioning databases and collections, you need to use the server API and authenticate via API key.

If you want to act as a single user, you need to authenticate using a session from your cookie jar, or set using the setSession helper.

021-server-side-rendering/021-server-side-rendering.md
loks0n: Love it 💣

021-server-side-rendering/021-server-side-rendering.md
gewenyu99: @loks0n sounds good to me.

021-server-side-rendering/021-server-side-rendering.md
```diff
@@ -0,0 +1,398 @@
...
+const sessionSecret = sessionCookie?.value;
+```
+
+> SDK for Flutter bypasses this problem by adding an 'interceptor' to requests. With any response containing a `set-cookie` header, the interceptor stores the cookies within a Flutter implementation of cookie storage. Before making any request, the session cookie is retrieved from storage and added to the request headers.
+
+#### Proposed Solution
+
+Return the token within the session object.
+
+Developers can then access the token like this:
```
stnguyen90, line 68: So for email + password session, the full flow is:

1. client sends email and password to backend
2. backend creates the session so that a secret is returned
3. backend sets a cookie so that the front end also has the session

Is that right?

loks0n: That's right

PR #60, opened 2024-03-07
Create 021-server-defined-attributes.md

vermakhushboo:
## What does this PR do?

RFC to add optional server-defined attributes to user 

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work.)

## Related PRs and Issues

(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

(Write your answer here.)

### DIFF
--- a/021-server-defined-attributes.md
+++ b/021-server-defined-attributes.md
@@ -1,0 +1,120 @@
+# Optional server-defined attributes
+
+* Creator: Khushboo Verma
+* Relevant Issues: N/A
+
+- [RFC Template](#rfc-template)
+  - [Summary](#summary)
+  - [Resources](#resources)
+  - [Implementation](#implementation)
+    - [API Changes](#api-changes)
+    - [Workers / Commands](#workers--commands)
+    - [Supporting Libraries](#supporting-libraries)
+    - [Data Structures](#data-structures)
+    - [SDKs](#sdks)
+    - [Breaking Changes](#breaking-changes)
+    - [Documentation & Content](#documentation--content)
+  - [Reliability](#reliability)
+    - [Security](#security)
+    - [Scaling](#scaling)
+    - [Benchmarks](#benchmarks)
+    - [Tests (UI, Unit, E2E)](#tests-ui-unit-e2e)
+  - [Open Questions](#open-questions)
+  - [Future Possibilities](#future-possibilities)
+
+## Summary
+<!-- Describe the problem we want to solve and suggested solution in a few paragraphs -->
+This RFC proposes the addition of optional server-defined attributes in all the project collections that can only be added by server and cannot be modified by the client. The user can add them to each collection through the API or console. Some useful sttributes can be `createdBy`, `updatedBy`, `ip` etc.
+
+## Resources
+<!-- List all the resources used for writing this RFC -->
+NA
+
+## Implementation
+
+<!-- Write an overview to explain the suggested implementation -->
+#### Console Changes
+We need to add an option in the `create attribute` settings dropdown to add a certain server-defined attribute from one of the existing options like `createdBy`.
+
+### API Changes
+<!-- Do we need new API endpoints? List and describe them and their API signatures -->
+We need 2 new API endpoints in `databases.php` for creating and deleting server-defined attributes.
+- POST `/v1/databases/:databaseId/collections/:collectionId/attributes/server`
+- DELETE `/v1/databases/:databaseId/collections/:collectionId/attributes/server/:key`
+
+We can allow users to create indexes for new attributes from the `createIndex` endpoint.
+
+We need to add a check in `createDocument` and `updateDocument` endpoints to populate the values of these attributes if they exist for client-side requests.
+
+###  Workers / Commands
+<!-- Do we need new workers or commands for this feature? List and describe them and their API signatures -->
+No new workers are anticipated.
+
+###  Supporting Libraries
+<!-- Do we need new libraries for this feature? Mention which, define the file structure, and different interfaces -->
+No new changes required in libraries.
+
+### Data Structures
+<!-- Do we need new data structures for this feature? Describe and explain the new collections and attributes -->
+#### In appwrite/appwrite
+The new attributes would add attributes to `attributes` collection in `$projectCollections` for each optional attribute.
+
+```
+$projectCollections = array_merge([
+  'attributes' => [
+        '$collection' => ID::custom(Database::METADATA),
+        '$id' => ID::custom('attributes'),
+        'name' => 'Attributes',
+        'attributes' => [
+            [
+                '$id' => ID::custom('createdBy'),
+                'type' => Database::VAR_STRING,
+                'format' => '',
+                'size' => 256,
+                'signed' => true,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => [],
+            ],
+...
+]
+```
+
+### SDKs
+<!-- Do we need to update our SDKs for this feature? Describe how -->
+SDKs will be updated to accommodate the changes in the API endpoints.
+
+### Breaking Changes
+<!-- Will this feature introduce any breaking changes? How can we achieve backward compatability -->
+No breaking changes are anticipated as this is an additive feature.
+
+### Documentation & Content
+<!-- What documentation do we need to update or add for this feature? -->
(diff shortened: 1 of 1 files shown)


021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
+# Optional server-defined attributes
+
+* Creator: Khushboo Verma
+* Relevant Issues: N/A
```
stnguyen90, line 4:
```suggestion
* Relevant Issues:
  * https://github.com/appwrite/appwrite/issues/3623
```

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+
+### API Changes
+<!-- Do we need new API endpoints? List and describe them and their API signatures -->
+We need 2 new API endpoints in `databases.php` for creating and deleting server-defined attributes.
+- POST `/v1/databases/:databaseId/collections/:collectionId/attributes/server`
+- DELETE `/v1/databases/:databaseId/collections/:collectionId/attributes/server/:key`
+
+We can allow users to create indexes for new attributes from the `createIndex` endpoint.
+
+We need to add a check in `createDocument` and `updateDocument` endpoints to populate the values of these attributes if they exist for client-side requests.
```
stnguyen90, line 47: We do updates at the utopia-php/database level as well so we might need to do it there instead. For example, updatedAt and createdAt are handled in utopia: https://github.com/utopia-php/database/blob/4199fe8f00f4e181c7782c4a6862845d591c1f03/src/Database/Database.php#L3050-L3051

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+
+###  Workers / Commands
+<!-- Do we need new workers or commands for this feature? List and describe them and their API signatures -->
+No new workers are anticipated.
+
+###  Supporting Libraries
+<!-- Do we need new libraries for this feature? Mention which, define the file structure, and different interfaces -->
+No new changes required in libraries.
```
Meldiron, line 55: Are you sure? Database library already auto-provides updatedAt, createdAt. These would also become server attributes, and it would feel weird to have some in Appwrite, and some in library.

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Resources
+<!-- List all the resources used for writing this RFC -->
+NA
+
+## Implementation
+
+<!-- Write an overview to explain the suggested implementation -->
+#### Console Changes
+We need to add an option in the `create attribute` settings dropdown to add a certain server-defined attribute from one of the existing options like `createdBy`.
```
Meldiron, line 37: Hmm I was imagining a toggle. Having it as option in create attribute, it sounds like it's attribute type. Its like I can have multiple, of the same type. But in reality I can only have one `$createdBy`. Also if attribute creation modal, it would feel weird that there is name input but its disabled.

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+
+### API Changes
+<!-- Do we need new API endpoints? List and describe them and their API signatures -->
+We need 2 new API endpoints in `databases.php` for creating and deleting server-defined attributes.
+- POST `/v1/databases/:databaseId/collections/:collectionId/attributes/server`
```
Meldiron, line 42: Lets also have endpoint for listing them, so we don't need to hard-code then on Console. Its better to have config for it in Appwrite or library.

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+]
+```
+
+### SDKs
+<!-- Do we need to update our SDKs for this feature? Describe how -->
+SDKs will be updated to accommodate the changes in the API endpoints.
+
+### Breaking Changes
+<!-- Will this feature introduce any breaking changes? How can we achieve backward compatability -->
+No breaking changes are anticipated as this is an additive feature.
```
Meldiron, line 90: We will need migration to enable $createdAt and $updatedAt on all collections made before the release of new version.

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+No new benchmarks are required as existing infrastructure will support the feature.
+
+### Tests (UI, Unit, E2E)
+<!-- How will we test this feature? -->
+- End-to-end testing for each new endpoint and functionality.
+- User Interface testing if a GUI component is added.
+
+## Open Questions
+<!-- List of things we need to figure out or farther discuss -->
+N/A
```
Meldiron, line 116: Which ones should we enable by default? None?

021-server-defined-attributes.md
Meldiron: I was also worried about design becasue dropdown is already pretty long, but Design team said dropdown will have search so design-wise its not a problem

021-server-defined-attributes.md
```diff
@@ -0,0 +1,120 @@
...
+    - [Security](#security)
+    - [Scaling](#scaling)
+    - [Benchmarks](#benchmarks)
+    - [Tests (UI, Unit, E2E)](#tests-ui-unit-e2e)
+  - [Open Questions](#open-questions)
+  - [Future Possibilities](#future-possibilities)
+
+## Summary
+<!-- Describe the problem we want to solve and suggested solution in a few paragraphs -->
+This RFC proposes the addition of optional server-defined attributes in all the project collections that can only be added by server and cannot be modified by the client. The user can add them to each collection through the API or console. Some useful sttributes can be `createdBy`, `updatedBy`, `ip` etc.
```
Meldiron, line 27: Let's have exact list of all internal attributes we will have

021-server-defined-attributes.md
vermakhushboo: Yes, resolving as discussed.

021-server-defined-attributes.md
vermakhushboo: Discussed as well. We won't allow enabling/ disabling attributes, so we don't need a toggle.

021-server-defined-attributes.md
vermakhushboo: Not relevant in the current approach.

021-server-defined-attributes.md
vermakhushboo: None by default. We add them when users create them.

byawitz:
The idea, in general, would help a lot on the part of Appwrite-console administrators (_That can be a new market title 😉_) and would help in setting pre-defined field structures for validations.

One way of expanding it would be to add a `default` callback using a function or other collection attributes.
This way the field value will auto-populate when no value was inserted.

**Example 1** - Next appointment date.
_In this example the default value would come from a function while passing the current document to that function_
```HTTP
POST https://cloud.appwrite/v1/databases/:databaseId/collections/:collectionId/attributes/server
Content-Type: application/json
origin: https://localhost
{
  "id"      : "next_appointment",
  "type"    : "datetime",
  "format"  : "",
  "size"    : 0,
  "required": false,
  "default" : "[FUNCTION_ID]",
  "array"   : false,
  "filters" : []
}
```

Then, the function would run when the default value is empty, something like 👇, for example.
```js

module.exports = async (context) => {
  try {
    // Set another appointment in 90 days.
    const next = new Date(+new Date() + (84600 * 90 * 1000));
    
    return context.res.json({ value: next });
  } catch (e) {
    return context.res.json({ value: '' });
  }
};
```

**Example 2** - Adding search attribute
_In this example the default value would come from the rest of the fields while each `%*%` variable is an attribute name in the current collection_
```HTTP
POST https://cloud.appwrite/v1/databases/:databaseId/collections/:collectionId/attributes/server
Content-Type: application/json
origin: https://localhost
{
  "id"      : "search",
  "type"    : "string",
  "format"  : "",
  "size"    : 512,
  "required": false,
  "default" : "%name% %email% %address% %country% %phone%",
  "array"   : false,
  "filters" : []
}
```

021-server-defined-attributes.md
stnguyen90: it would also be good to clarify exactly what the behavior of each would be.

jasonetorres:
👀👀🔥 definitely checking this out

PR #61, opened 2024-03-21
Create 021-presence-service

eldadfux:
Parsed version: https://github.com/appwrite/rfc/blob/a60c657ce2b8580c6c51013be509ccda690b693d/021-presence-service.md


021-presence-service.md
```diff
@@ -0,0 +1,159 @@
+# Presence service
+
+* Creator: Eldad Fux
+* Relevant Issues:
```
stnguyen90, line 4: Related: https://github.com/appwrite/appwrite/issues/2418

021-presence-service.md
```diff
@@ -0,0 +1,159 @@
...
+<!-- Write an overview to explain the suggested implementation -->
+
+### API Changes
+<!-- Do we need new API endpoints? List and describe them and their API signatures -->
+
+```POST /v1/presences```
+
+Create a presence. Set the user presence status, expiry (minutes/date - TBD), and permissions for who has access to see the current presence. This will imitate the equivalent action in the GraphQL and Realtime APIs.
+
+```GET /v1/presences```
```
Meldiron, line 46: Lets please add response model. Im curious if there will be connection between presence ID and user ID, and if its public.

021-presence-service.md
```diff
@@ -0,0 +1,159 @@
...
+
+List all presences, on client side this will show only presneces I have permission to view. On server it will show all presences on the server. This endpoint will support queries and pagination like any other list endpoint on Appwrite.
```
Meldiron:
```suggestion
List all presences, on client side this will show only presneces I have permission to view. On server it will show all presences. This endpoint will support queries and pagination like any other list endpoint on Appwrite.
```

Meldiron: Hmm, I think we should have a shortcut for getting "my presences". I don't know how - I cant see good way. But I would still like it.

eldadfux: Good point.

eldadfux: We could use a query filter with user id, but this is not as quick.

021-presence-service.md
Meldiron: I am also curious what data can I set on presence.. message, icon, color? picture, age, status?

021-presence-service.md
```diff
@@ -0,0 +1,163 @@
...
+Delete presence by ID.
+
+### Events
+
+The new endpoints will trigger standard events that will be used for listening to Realtime API changes or to execute functions and webhooks.
+
+### Console updates
+
+- We should implement presence in the console itself and make current org members online or not or last online
+- We will add a new section under auth called `Presences`, this section will show a dashboard with usage showing how many people were online in a specific time range. Below that, we will show avatars of all currently online users or users who just recently went offline.
```
Meldiron, line 69: Hmm I imagined this as tab under user detail page. Can 1 user have multiple presences?

021-presence-service.md
```diff
@@ -0,0 +1,163 @@
...
+We need to add a new collection to store presences.
+
+- _id
+- _uid
+- _createdAt
+- _updatedAt
+- userInternalId
+- userId
+- expiry (DateTime)
+- status (string)
```
Meldiron, line 98: I think this should be JSON, since often apps need more than just a message. GitHub lets you set emoji. Discord lets you also set game you currently play. Instagram lets you ass music to your note.

021-presence-service.md
```diff
@@ -0,0 +1,163 @@
...
+
+We need to add a new collection to store presences.
+
+- _id
+- _uid
+- _createdAt
+- _updatedAt
+- userInternalId
+- userId
+- expiry (DateTime)
```
Meldiron, line 97: Will exipry be nullable - never delete?

021-presence-service.md
ArnabChatterjee20k: separate metadata and not storing inside the status itself as User can query by status as well

And having json on status can be performance bottleneck(mysql)

Having metadata allows user to have fancy status like we have in github, discord ,etc

021-presence-service.md
ArnabChatterjee20k: I think we can't allow a big expiry. There should be max value otherwise the stale presence count will be high. Max 15mins/1hour?

